PR #10, opened 2021-03-01
Work in progress, removing euler from the protocol, providing util functions

samcake:


src/classes/HiFiAudioAPIData.ts
```diff
@@ -33,6 +33,30 @@ export class Point3D {
...
+export class OrientationQuat3D {
+    w: number;
+    x: number;
+    y: number;
+    z: number;
+
+    /**
+     * Construct a new `OrientationQuat3D` object. All parameters are required.
+     */
+    constructor({ w = 1, x = 0, y = 0, z = 0 }: { w?: number, x?: number, y?: number, z?: number } = {}) {
```
zfox23, line 51: this isn't the same as what's in my branch, and i think this is going to cause problems.

if the last sent quat data component "x" is the same as the current quat data component "x", we won't send the "x" component of the quat up to the server.

check out the diff function on line 248 below. maybe you can fix it there and keep this as the constructor (i like it better this way, if we can make it work)

zfox23 (changes requested):

src/classes/HiFiCommunicator.ts
```diff
@@ -428,30 +427,15 @@ export class HiFiCommunicator {
             this._currentHiFiAudioAPIData.position.z = position.z ?? this._currentHiFiAudioAPIData.position.z;
         }
 
-        if (orientationEuler) {
-            if (!this._currentHiFiAudioAPIData.orientationEuler) {
-                this._currentHiFiAudioAPIData.orientationEuler = new OrientationEuler3D();
+        if (orientation) {
+            if (this._currentHiFiAudioAPIData.orientation) {
```
zfox23, line 431:
```suggestion
            if (!this._currentHiFiAudioAPIData.orientation) {
```

src/classes/HiFiAxisConfiguration.ts
```diff
@@ -217,36 +217,41 @@ export class HiFiAxisUtilities {
...
+    }
 
+    /**
+     * ⚠ WARNING ⚠ The code in this function IS be wrong.
+     * TODO: implement the function, just a NO OP at the moment.
+     * 
+     * @param axisConfiguration 
+     * @param inputOrientationQuat3D 
+     */
+    static translateOrientation3FromMixerSpace(axisConfiguration: HiFiAxisConfiguration, mixerOrientationQuat3D: OrientationQuat3D): OrientationQuat3D {
```
zfox23, line 252:
```suggestion
    static translateOrientationQuat3DFromMixerSpace(axisConfiguration: HiFiAxisConfiguration, mixerOrientationQuat3D: OrientationQuat3D): OrientationQuat3D {
```

src/classes/HiFiAxisConfiguration.ts
```diff
@@ -217,36 +217,41 @@ export class HiFiAxisUtilities {
...
-            retval.rollDegrees = inputOrientationEuler3D.rollDegrees;
-        }
+    /**
+     * ⚠ WARNING ⚠ The code in this function IS be wrong.
+     * TODO: implement the function, just a NO OP at the moment.
+     * 
+     * @param axisConfiguration 
+     * @param inputOrientationQuat3D 
+     */
+    static translateOrientation3DToMixerSpace(axisConfiguration: HiFiAxisConfiguration, inputOrientationQuat3D: OrientationQuat3D): OrientationQuat3D {
```
zfox23, line 239:
```suggestion
    static translateOrientationQuat3DToMixerSpace(axisConfiguration: HiFiAxisConfiguration, inputOrientationQuat3D: OrientationQuat3D): OrientationQuat3D {
```

src/classes/HiFiAudioAPIData.ts
samcake: I addressed the diff feature dierectly in the mixer, and sowe don;t need the geometrical structs to be able to have "null" components.
As discussed we prefer point, quat or even euler to have their componetns fully defined

src/classes/HiFiAudioAPIData.ts
```diff
@@ -66,33 +101,249 @@ export class OrientationEuler3D {
      * Construct a new `OrientationEuler3D` object. All parameters are optional. Unset parameters will be set to `0`. Remember, all units for member variables are `degrees`.
      */
     constructor({ pitchDegrees = 0, yawDegrees = 0, rollDegrees = 0 }: { pitchDegrees?: number, yawDegrees?: number, rollDegrees?: number } = {}) {
-        this.pitchDegrees = pitchDegrees;
-        this.yawDegrees = yawDegrees;
-        this.rollDegrees = rollDegrees;
+        this.pitchDegrees = clampNonan(pitchDegrees, -360, 360, 0);
```
zfox23, line 104: why do we want to _clamp_ these values? why wouldn't we want to add/subtract 360 degrees to ensure the values are within range?

src/classes/HiFiAudioAPIData.ts
```diff
@@ -66,33 +101,249 @@ export class OrientationEuler3D {
...
+ *  2/ Pitch, rotating around the right axis 
+ *  3/ Roll, rotating around the front axis
+ *  the resulting 3d frame orientation is relative to the base frame.
  */
-export class OrientationQuat3D {
-    w: number;
-    x: number;
-    y: number;
-    z: number
+export enum OrientationEuler3DOrder {
```
zfox23, line 122: this is great, thank you.

src/classes/HiFiAudioAPIData.ts
```diff
@@ -66,33 +101,249 @@ export class OrientationEuler3D {
...
+
+    // then depending on the euler rotation order decomposition, we extract the angles 
+    // from the base vector components
+    let pitch = 0;
+    let yaw = 0;
+    let roll = 0;
+    switch (order) {
+    case OrientationEuler3DOrder.PitchYawRoll: {
+        yaw = Math.asin( r02 );
+        if ( Math.abs( r02 ) < 0.9999999 ) {
```
zfox23, line 292: any reason not to do this here and below?:

```suggestion
        if ( Math.abs( r02 ) < 1 - Number.EPSILON ) {
```

src/classes/HiFiAudioAPIData.ts
```diff
@@ -280,7 +464,7 @@ export class ReceivedHiFiAudioAPIData extends HiFiAudioAPIData {
      * ❌ The client never sends `volumeDecibels` data to the server.
      * ✔ The server sends `volumeDecibels` data to all clients connected to a server during "peer updates".
      */
-    constructor(params: { providedUserID?: string, hashedVisitID?: string, volumeDecibels?: number, position?: Point3D, orientationEuler?: OrientationEuler3D, orientationQuat?: OrientationQuat3D, hiFiGain?: number } = {}) {
+    constructor(params: { providedUserID?: string, hashedVisitID?: string, volumeDecibels?: number, position?: Point3D, orientation?: OrientationQuat3D, hiFiGain?: number } = {}) {
```
zfox23, line 467: this one seems like it might be confusing for developers.

i vote keep the `orientationQuat` name in here so folks aren't confused about whether or not the received orientation is quat vs euler.

this might not be the only place where keeping the old name would be helpful, it's just the one i noticed first

src/classes/HiFiAxisConfiguration.ts
```diff
@@ -43,6 +45,7 @@ export class HiFiAxisConfiguration {
  * - `+y` is into the screen and `-y` is out of the screen towards the user
```
zfox23, line 45: this old comment is wrong and needs to be changed, correct?

src/classes/HiFiAxisConfiguration.ts
```diff
@@ -217,36 +221,41 @@ export class HiFiAxisUtilities {
     }
 
     /**
-     * ⚠ WARNING ⚠ The code in this function might be wrong, because 3D math is really hard. The default configuration works fine,
-     * but it's challenging to verify that other configurations work as expected until we have a better 3D example app.
-     * TODO: Verify that this is actually doing what we want for it to be doing.
+     * ⚠ WARNING ⚠ The code in this function IS wrong.
```
zfox23, line 224: i assume you're still workin' on this and below :)

src/classes/HiFiCommunicator.ts
```diff
@@ -153,6 +153,9 @@ export class HiFiCommunicator {
                 HiFiLogger.error(`There is an error with the passed \`HiFiAxisConfiguration\`, so the new axis configuration was not set. There are more error details in the logs above.`);
             }
         }
+
+        // initialize the current Audo API Data with the given data, but use the updateUserData call for sanity.
```
zfox23, line 157:
```suggestion
        // Initialize the current Audio API Data with the given data, but use the `updateUserData()` call for sanity.
```

src/classes/HiFiMixerSession.ts
```diff
@@ -660,7 +663,7 @@ export class HiFiMixerSession {
      * @returns If this operation is successful, returns `{ success: true, stringifiedDataForMixer: <the raw data that was transmitted to the server>}`. If unsuccessful, returns
      * `{ success: false, error: <an error message> }`.
      */
-    _transmitHiFiAudioAPIDataToServer(hifiAudioAPIData: HiFiAudioAPIData): any {
+    _transmitHiFiAudioAPIDataToServer(currentHifiAudioAPIData: HiFiAudioAPIData, previousHifiAudioAPIData?: HiFiAudioAPIData): any {
```
zfox23, line 666: i _think_ what you've done below is make the diff function unused. is that right? if so, is there any reason to keep the diff function? also, if that's right, you'll need to add some new properties here (volume threshold, etc)

zfox23 (changes requested):

src/classes/HiFiAudioAPIData.ts
samcake: sounds good

src/classes/HiFiMixerSession.ts
samcake: Correct,
the diff function has been removed yes
and the diff is happening directly in this function comparing the last data sent with the new one

src/classes/HiFiAudioAPIData.ts
```diff
@@ -193,83 +446,6 @@ export class HiFiAudioAPIData {
         this.userAttenuation = userAttenuation;
         this.userRolloff = userRolloff;
     }
-
```
samcake, line 196: No more DIff function, the diff is happening directly when it is time to send a new payload to the server

luiscuenca (approved):

src/classes/HiFiAudioAPIData.ts
```diff
@@ -120,16 +387,17 @@ export class HiFiAudioAPIData {
...
-     * 
-     * ✔ The client sends `orientationEuler` data to the server when `_transmitHiFiAudioAPIDataToServer()` is called.
-     * 
-     * ✔ The server sends `orientationEuler` data to all clients connected to a server during "peer updates".
      * @param orientationQuat If you don't supply an `orientationQuat` when constructing instantiations of this class, `orientationQuat` will be `null`.
      * 
      * ✔ The client sends `orientationQuat` data to the server when `_transmitHiFiAudioAPIDataToServer()` is called.
      * 
      * ✔ The server sends `orientationQuat` data to all clients connected to a server during "peer updates".
+     * @param orientationEuler For convenience, a euler representation of the orientation is supported.
```
zfox23, line 395:
```suggestion
     * @param orientationEuler For convenience, a Euler representation of the orientation is supported.
```

src/classes/HiFiAudioAPIData.ts
```diff
@@ -66,33 +116,250 @@ export class OrientationEuler3D {
...
+ *  The resulting rotation is defining the 'rotated' space relative to the 'base' space.
+ *  A vector Vr in "rotated' space and its equivalent value Vb in the'base' space is computed as follow:
+ *  Vb = [P][Y][R] Vr
+ * 
+ * @param quat - The orientation quaternion.
+ * @param order - The euler order convention.
+ * 
+ * @return The end resulting quaternion defined from the euler angles combination
+ */
+export function eulerFromQuaternion(quat: OrientationQuat3D, order : OrientationEuler3DOrder): OrientationEuler3D {
```
zfox23, line 272:
```suggestion
export function eulerFromQuaternion(quat: OrientationQuat3D, order: OrientationEuler3DOrder): OrientationEuler3D {
```

src/classes/HiFiAxisConfiguration.ts
```diff
@@ -32,17 +32,20 @@ export class HiFiAxisConfiguration {
...
 
 /**
  * Contains the application's 3D axis configuration. By default:
  * - `+x` is to the right and `-x` is to the left
- * - `+y` is into the screen and `-y` is out of the screen towards the user
- * - `+z` is up and `-z` is down
+ * - `+y` is up and `-z` is down
+ * - `+z` is back and `-z` is front
  * - The coordinate system is right-handed.
+ * - euler order is YawPitchRoll
```
zfox23, line 48:
```suggestion
 * - Euler order is `OrientationEuler3DOrder.YawPitchRoll`
```

src/classes/HiFiCommunicator.ts
```diff
@@ -360,10 +363,17 @@ export class HiFiCommunicator {
      * the user data on the High Fidelity Audio API server. There are no good reasons for a client to call this function
      * and _not_ update the server User Data, and thus this function is `private`.
      * 
+     * The function both representation of orientation, quaternion and euler.
```
zfox23, line 366:
```suggestion
     * You can update user orientation by passing Quaternion or Euler orientation representations to this function.
```

zfox23 (changes requested):

luiscuenca (approved):

zfox23 (approved):

PR #11, opened 2021-03-02
HIFI-358: Adding ability to set individual attenuation and rolloff per-connection

hd23408:
New `userAttenuation` and `userRolloff` settings on HiFiAudioAPIData.ts, plus documentation. Requires running v1.26.2 on the server side.

(Currently working on adding an example of usage to the "complex" sample app as well.)

### DIFF
--- a/src/classes/HiFiAudioAPIData.ts
+++ b/src/classes/HiFiAudioAPIData.ts
@@ -165,7 +165,7 @@ export class HiFiAudioAPIData {
      * ❌ The server never sends `userAttenuation` data.
      *
      * @param userRolloff This value represents the progressive high frequency roll-off in meters, a measure of how the higher frequencies 
-     * in a user's sound are dampened as the user gets further away. By default, there is a global roll-off value (set for a given space), currently 12.5 
+     * in a user's sound are dampened as the user gets further away. By default, there is a global roll-off value (set for a given space), currently 16 
      * meters, which applies to all users in a space. This value represents the distance for a 1kHz rolloff. Values in the range of 
      * 12 to 32 meters provide a more "enclosed" sound, in which high frequencies tend to be dampened over distance as they are 
      * in the real world. Generally changes to roll-off values should be made for the entire space rather than for individual users, but
--- a/src/classes/HiFiAudioAPIData.ts
+++ b/src/classes/HiFiAudioAPIData.ts
@@ -140,28 +140,37 @@ export class HiFiAudioAPIData {
      * ✔ The server sends `hiFiGain` data to all clients connected to a server during "peer updates".
 
      * @param userAttenuation This value affects how far a user's sound will travel in 3D space, without affecting the user's loudness.
-     * By default, there is a global attenuation value (set for a given space) that applies to all users in a space.
+     * By default, there is a global attenuation value (set for a given space) that applies to all users in a space. This default space
+     * attenuation is usually 0.5, which represents a reasonable approximation of a real-world fall-off in sound over distance.
+     * Lower numbers represent less attenuation (i.e. sound travels farther); higher numbers represent more attenuation (i.e. sound drops
+     * off more quickly).
+     * 
+     * When setting this value for an individual user, the following holds:
+     *   - Positive numbers should be between 0 and 1, and they represent a logarithmic attenuation. This range is recommended, as it is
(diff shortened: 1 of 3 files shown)


src/classes/HiFiAudioAPIData.ts
```diff
@@ -135,12 +138,52 @@ export class HiFiAudioAPIData {
...
+     * 
+     * When setting this value for an individual user, the following holds:
+     *   - Positive numbers should be between 0 and 1, and they represent a logarithmic attenuation. This range is recommended, as it is
+     * more natural sounding.  Smaller numbers represent less attenuation, so a number such as 0.2 can be used to make a particular 
+     * user's audio travel farther than other users', for instance in "amplified" concert type settings. Similarly, an extremely 
+     * small non-zero number (e.g. 0.00001) can be used to effectively turn off attenuation for a given user within a reasonably 
+     * sized space, resulting in a "broadcast mode" where the user can be heard throughout most of the space regardless of their location
+     * relative to other users. (Note: The actual value "0" is used internally to represent the default; for setting minimal attenuation, 
+     * small non-zero numbers should be used instead. See also "userRolloff" below.)
+     *   - Negative attenuation numbers are used to represent linear attenuation, and are a somewhat artificial, non-real-world concept. However,
```
kencooke, line 156: Might want to point out that for linear attenuation, the setting is the distance (meters) where the audio becomes totally inaudible..

src/classes/HiFiAudioAPIData.ts
```diff
@@ -135,12 +138,52 @@ export class HiFiAudioAPIData {
...
+     *
+     * If you don't supply an `userAttenuation` when constructing instantiations of this class, `userAttenuation` will be `null` and the
+     * default will be used.
+     * 
+     * ✔ The client sends `userAttenuation` data to the server when `_transmitHiFiAudioAPIDataToServer()` is called.
+     * 
+     * ❌ The server never sends `userAttenuation` data.
+     *
+     * @param userRolloff This value represents the progressive high frequency roll-off in meters, a measure of how the higher frequencies 
+     * in a user's sound are dampened as the user gets further away. By default, there is a global roll-off value (set for a given space), currently 12.5
```
kencooke, line 167: I think we just changed the default to 16m?

src/classes/HiFiAudioAPIData.ts
hd23408: Thank you, @kencooke - will do!

src/classes/HiFiAudioAPIData.ts
hd23408: Yeah, I think @howard-stearns is planning to make that change but maybe hasn't yet? I'll update the docs when the mixer gets updated.

zfox23 (dismissed):

src/classes/HiFiAudioAPIData.ts
howard-stearns: Roxanne did it during a break. Not in v1.26.2. https://github.com/highfidelity/audionet-hifi/pull/268/files

src/classes/HiFiCommunicator.ts
```diff
@@ -457,6 +461,13 @@ export class HiFiCommunicator {
         if (typeof (hiFiGain) === "number") {
             this._currentHiFiAudioAPIData.hiFiGain = Math.max(0, hiFiGain);
         }
+        if (typeof (userAttenuation) === "number") {
+            this._currentHiFiAudioAPIData.userAttenuation = userAttenuation;
+        }
+        if (typeof (userRolloff) === "number") {
+            // TODO: will this ever be negative legitimately?
```
howard-stearns, line 468: In my opinion, this style of type checking is not helpful to the user (as it mysteriously changes the user's data without warning,) doesn't really deliver any safety to either users or server, and makes the code harder to maintain, extend, and audit.

However, GIVEN that we are doing this anyway, AND that we have declared that (until we change our minds) we will not support negative values, then I think the answer to the question is that it will not legitimately be negative and it would be consistent with the existing code here to block negative values.

src/classes/HiFiAudioAPIData.ts
hd23408: now in v1.26.3 though, correct @howard-stearns ? I'll update the comment.

howard-stearns: correct

PR #12, opened 2021-03-02
Replace stackName from URL parameters with hostURL

luiscuenca:
This PR removes all logic that extract the **webSocket** address from the `stackName` URL param. Now this information is passed using a `hostURL` function parameter.


src/classes/HiFiCommunicator.ts
```diff
@@ -187,24 +187,14 @@ export class HiFiCommunicator {
...
-     * it is very likely that we are running a test in a browser. Tests running in a browser from that hostname should assume that the WebRTC Signaling Address
-     * is at the same host from which the test is served.
-     * 3. If the code is running in the browser context, and our code compilation processes have specified the build mode as "production", we should use the production
-     * WebRTC signaling connection address.
-     * 4. If a developer has passed a `stackName` parameter into this `connectToHiFiAudioAPIServer()` call, use a WebRTC signaling address based on that `stackName`.
-     *     - If the passed `stackName` is a URL, that URL will be used as the WebRTC Signaling Address directly.
-     * 5. If the code is running in the NodeJS context, we will use the "production" `stackName`, and use a WebRTC signaling address based on that `stackName`.
-     * 6. If none of the above logic applies, we will use the default "staging" WebRTC signaling connection address.
+     * @param hostURL An URL that will be used to create a valid WebRTC signaling address. `wws://${hostURL}:8001/?token=`
```
zfox23, line 190:
```suggestion
     * @param hostURL An URL that will be used to create a valid WebRTC signaling address. The passed `hostURL` will be used to construct a signaling address of the form: `wss://${hostURL}:8001/?token=`
```

src/classes/HiFiCommunicator.ts
```diff
@@ -187,24 +187,14 @@ export class HiFiCommunicator {
...
+     * If the developer does not pass a hostURL parameter, a default URL will be used instead. HiFiConstants.DEFAULT_PROD_HIGH_FIDELITY_ENDPOINT
```
zfox23, line 191:
```suggestion
     * If the developer does not pass a `hostURL` parameter, a default URL will be used instead: See: {@link DEFAULT_PROD_HIGH_FIDELITY_ENDPOINT}
```

src/classes/HiFiCommunicator.ts
```diff
@@ -187,24 +187,14 @@ export class HiFiCommunicator {
...
+     * Reading this parameter from the URL should be implemented by the developer as part of the application code.
```
zfox23, line 192: What do you mean by this?

src/classes/HiFiCommunicator.ts
```diff
@@ -214,56 +204,17 @@ export class HiFiCommunicator {
...
-                }
-            } else if (!isBrowserContext) {
-                webRTCSignalingAddress = `${HiFiConstants.DEFAULT_PROD_HIGH_FIDELITY_ENDPOINT}/?token=`;
-            } else {
-                webRTCSignalingAddress = webRTCSignalingAddress.replace('$STACKNAME', 'api-staging-02');
-            }
+            hostURLSafe = new URL(hostURL).hostname;
+        } catch(e) {
+            // If hostURL is not defined, we assign the default URL
+            hostURLSafe = hostURL ? hostURL : HiFiConstants.DEFAULT_PROD_HIGH_FIDELITY_ENDPOINT;
```
zfox23, line 213: why wouldn't we want to do this:
```suggestion
            hostURLSafe = HiFi4Constants.DEFAULT_PROD_HIGH_FIDELITY_ENDPOINT;
```
?

zfox23 (changes requested):

src/classes/HiFiCommunicator.ts
luiscuenca: This is a note that makes reference to this change. Basically we say that we are not getting information from the URL params anymore. Now we rely on what URL is passed during the connection call. This means that it is the application itself that needs to get this address from the URL params if necessary, so the developer is now in charge of implementing this.

src/classes/HiFiCommunicator.ts
luiscuenca: The first thing we do is trying to catch a complete URL (i.e. `https://api.highfidelity.com/`). If that fails we assume that either `hostURL` is undefined or it is not a complete URL (i.e `api.highfidelity.com`). Keep in mind that `new URL("api.highfidelity.com")` will fail.

src/classes/HiFiCommunicator.ts
luiscuenca: Fixed

src/classes/HiFiCommunicator.ts
luiscuenca: Fixed

src/classes/HiFiCommunicator.ts
zfox23: oh okay i see!

src/classes/HiFiCommunicator.ts
zfox23: aha! okay, thanks for the explanation.

zfox23 (approved):

src/classes/HiFiCommunicator.ts
hd23408: @luiscuenca Sorry, late to the party here -- does that mean that when a dev passes in "hostURL" it needs to be a complete URL of the format `https://whatever.highfidelity.com/` and NOT just `whatever.highfidelity.com`? I had interpreted from the "`wws://${hostURL}:8001/?token=`" comment you had above that `whatever.highfidelity.com` was the right format. Can you clarify in the comments exactly which format is expected? Thank you!!

PR #14, opened 2021-03-03
Add volume threshold

howard-stearns:
https://highfidelity.atlassian.net/browse/HIFI-357

### DIFF
--- a/src/classes/HiFiAudioAPIData.ts
+++ b/src/classes/HiFiAudioAPIData.ts
@@ -107,6 +107,7 @@ export class HiFiAudioAPIData {
     position: Point3D;
     orientationEuler: OrientationEuler3D;
     orientationQuat: OrientationQuat3D;
+    volumeThreshold: number;
     hiFiGain: number;
 
     /**
@@ -127,6 +128,10 @@ export class HiFiAudioAPIData {
      * ✔ The client sends `orientationQuat` data to the server when `_transmitHiFiAudioAPIDataToServer()` is called.
      * 
      * ✔ The server sends `orientationQuat` data to all clients connected to a server during "peer updates".
+     *
+     * @param volumeThreshold A volume level below this value is considered background noise and will be smoothly gated off.
+     * The floating point value is in the same decibel units as the VolumeDecibels component of UserDataSubscription.
+     *
      * @param hiFiGain This value affects how loud User A will sound to User B at a given distance in 3D space.
      * This value also affects the distance at which User A can be heard in 3D space.
      * Higher values for User A means that User A will sound louder to other users around User A, and it also means that User A will be audible from a greater distance.
@@ -136,10 +141,11 @@ export class HiFiAudioAPIData {
      * 
      * ✔ The server sends `hiFiGain` data to all clients connected to a server during "peer updates".
      */
-    constructor({ position = null, orientationEuler = null, orientationQuat = null, hiFiGain = null }: { position?: Point3D, orientationEuler?: OrientationEuler3D, orientationQuat?: OrientationQuat3D, hiFiGain?: number } = {}) {
+    constructor({ position = null, orientationEuler = null, orientationQuat = null, volumeThreshold = null, hiFiGain = null }: { position?: Point3D, orientationEuler?: OrientationEuler3D, orientationQuat?: OrientationQuat3D, volumeThreshold?: number, hiFiGain?: number } = {}) {
         this.position = position;
         this.orientationQuat = orientationQuat;
         this.orientationEuler = orientationEuler;
+        this.volumeThreshold = volumeThreshold;
         this.hiFiGain = hiFiGain;
     }
 
@@ -154,6 +160,9 @@ export class HiFiAudioAPIData {
             "orientationEuler": Object.assign({}, this.orientationEuler),
             "orientationQuat": Object.assign({}, this.orientationQuat),
         };
+        if (typeof (this.volumeThreshold) === "number") {
+            currentHiFiAudioAPIDataObj["volumeThreshold"] = this.volumeThreshold;
+        }
         if (typeof (this.hiFiGain) === "number") {
             currentHiFiAudioAPIDataObj["hiFiGain"] = this.hiFiGain;
         }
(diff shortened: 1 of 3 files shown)


src/classes/HiFiAudioAPIData.ts
```diff
@@ -127,6 +128,10 @@ export class HiFiAudioAPIData {
      * ✔ The client sends `orientationQuat` data to the server when `_transmitHiFiAudioAPIDataToServer()` is called.
      * 
      * ✔ The server sends `orientationQuat` data to all clients connected to a server during "peer updates".
+     *
+     * @param volumeThreshold A volume level below this value is considered background noise and will be smoothly gated off.
+     * The floating point value is in the same decibel units as the VolumeDecibels component of UserDataSubscription.
```
kencooke, line 133: To prevent  people from using positive numbers here, might say something like `volumeThreshold is specified in dBFS (decibels relative to full scale) with values between -96dB (disabled) and 0dB.`

howard-stearns: Done.  

Question though, internally: I believe that I've seen positive numbers come back from RaviAudioStream::getPreGateLevel(), but I might be mistaken.  Would this mean that something is wrong, or is it that it isn't particularly useful to set the volumeThreshold above zero?

src/classes/HiFiAudioAPIData.ts
```diff
@@ -107,6 +107,7 @@ export class HiFiAudioAPIData {
     position: Point3D;
     orientationEuler: OrientationEuler3D;
     orientationQuat: OrientationQuat3D;
+    volumeThreshold: number;
```
hd23408, line 110: I'm not in love with this variable name, but I'm having trouble coming up with a better one. :-) I immediately assumed that `volumeThreshold` was the MAXIMUM volume that a user connection could have, rather than (effectively) the minimum volume. Could this be considered to be the `noiseFloor`? Would that be confusing? (cc @kencooke )

howard-stearns: FYI and FWIW, not being immersed in the field: 
- To me naively, "limit" could be either floor or ceiling, but to refer to a maximum or ceiling I would use "limit", rather than "threshold". (I would take threshold to mean a floor or minimum accepted value.)
- Threshold is the term used in the mixer and the gate, although the word "volume" is not used. My thinking (such as it is) in adding the term "volume" is to pair it with the units that come back from the volume subscription, rather than -1 to 1, or a 16 bit int, etc.
- To me, a "noise floor" would be the lowest amount of noise one can hear. We are talking here about a "signal floor", or a "noise ceiling". (But alas, I don't think that "noise ceiling" would be very clear either. Argh.)

hd23408: Maybe "lowVolumeThreshold"? I definitely don't feel strongly here -- you do a good job of explaining it in the comments.

kencooke: Well, there is a loooooong history (in audio recording/broadcasting industry) of the activation point of a dynamic range processing tool being called a gate/compressor/limiter `threshold`. I'd prefer to explain the threshold in layman terms, than to make up a new and unconventional name for it.

src/classes/HiFiAudioAPIData.ts
kencooke: The complication is we have a 16-bit mic stream, and a gain modifier that can amplify/attenuate this into HDR audio. The gate threshold only relates to recording properly, and [-96,0] covers the entire range at this point. Any metering used to set recording levels and threshold should also be raw mic level. Once we apply the gain,  levels can be much greater (up to +96dB) to the rest of the system. For example, this is the level the mixer would consider if ranking streams by input volume for throttling purposes.

src/classes/HiFiAudioAPIData.ts
howard-stearns: I don't have a preference between volumeThreshold, lowVolumeThreshold, or volumeGateThreshold. I'm good with any of them.

src/classes/HiFiAudioAPIData.ts
howard-stearns: Does that mean that RaviAudioStream::getPreGateLevel() should only be between -96 and 0, or that only AudioGate::getPeakLevel() should be between -96 and 0 (or neither, and I'm still lost in the sea of misunderstanding)?

We (now) report the former to the client subscription (RaviAudioStream::getPreGateLevel()). I want to make sure that we're reporting the right thing to the user, so that they could implement "Ken's Pro Limiter UI" if they wanted to.

(It is possible that my memory of positive numbers at the client was from an earlier version of The Sound of Silence PR, BEFORE I split the report gate level into a RaviAudioStream::getPreGateLevel() and a getPostGateLevel(), such the positive numbers were from what is now called getPostGateLevel().)

kencooke: I think the key here is to name/explain them to encourage the intended usage.
For example, `setMicThreshold` and `getMicLevel` as a pair intended to set recording level properly.
In this case, the level would NOT include the post-gain.
Then something like 'getInputLevel` or `getUserLevel` could report the post-gate, post-gain level that everyone  else actually hears.

howard-stearns: This is complicated stuff, and I've probably over-simplified too much. I'd like to get everything correct. I'm not certain what our requirements should be, but I'm assuming:

1. We need to publish a "volume" of some sort to the user, and we can decide, change, and document the specific behavior.

2. We should not publish an additional (second) stream of volume data beyond the current one. (There's plenty to argue about there -- and I've certainly tried -- but the bottom line is that we have what we have in terms of architecture and features, and no one is going to be happy if we say that we have to reduce our rated capacities.)

3. For that published volume to be most useful, I (think?) it should be independent of other client-settable parameters such as "hifiGain", "userAttenuation", and whatever we call this gating threshold, so that setting the other parameters doesn't effect the the published value. (Boy, our names really are all over the place!)

4. I (think?) app developers will have less confusion than otherwise if the units/scale of the published volume are the same as that for this threshold.

5. We currently have many different kinds of volumes internally available. If we can come up with one that meets, eg., 3 and 4, then I would like to use that as the one "volume" that we publish.

Some possibilities:
- Maybe AudioGate::getPeakLevel() does it?
- Maybe I've got the wrong idea with requirement 3, e.g., maybe volume should reflect "hifiGain". 
- Maybe in requirement 2, it's fine for the information about EVERYONEs volume to be one thing, and that we would do something fancy and specific for setting your threshold. E.g., we don't currently have capability to temporarily receive additional data about yourself.
- Maybe we're going overboard with allowing users to set this threshold, and should just publicly have gate on/off, or music-mode?

zfox23 (approved):

zfox23:
Approved, but there's a merge conflict!

zfox23:
i fixed the merge conflict, and then merged :)

PR #15, opened 2021-03-03
Test to check space IDs

RebeccaStankus:


tests/integration/rest.integration.test.ts
```diff
@@ -74,6 +74,31 @@ describe('HiFi API REST Calls', () => {
...
+
+        test(`the list is accurate`, async () => {
+            let returnMessage = await fetch(`${stackURL}/api/v1/spaces/?token=${adminToken}`);
+
+            let spacesListJSON: any = {};
+            spacesListJSON = await returnMessage.json();
+            expect(spacesListJSON.length).toBe(Object.keys(stackData.apps.app1.spaces).length);
+            spacesListJSON.forEach(async (space: any) => {
+                let match = false;
+                for (var key in stackData.apps.app1.spaces) {
```
zfox23, line 94:
```suggestion
                for (let key in stackData.apps.app1.spaces) {
```

tests/integration/rest.integration.test.ts
```diff
@@ -14,7 +14,7 @@ describe('HiFi API REST Calls', () => {
             nonAdminToken = await generateJWT(TOKEN_GEN_TYPES.NON_ADMIN_ID_APP1_SPACE1_SIGNED);
         } catch (err) {
             console.error("Unable to create non admin token for testing REST calls. ERR: ", err);
-            return;
+            process.exit();
```
zfox23, line 17: is this something that needs to happen after every test or just this one?

zfox23 (changes requested):

tests/integration/rest.integration.test.ts
RebeccaStankus: This only needs to happen here if we can't set up for the tests by creating the admin token. I've actually tried adding it after every test to fix the 'worker process has failed to exit...' bug but then the test suite stops completely after the first test. I can only call this once per file and then the script stops.

zfox23: the first sentence makes sense to me

the latter part of your explanation also makes sense - if you call `process.exit()` at any time, that entire test suite will stop at the time that function is called.

tests/integration/secrets/auth_example.json
```diff
@@ -2,29 +2,47 @@
     "stackData": {
         "name": "<YOUR STACK NAME>",
         "url": "<YOUR STACK URL>",
+        "wss": "<YOUR STACK WSS URL>",
```
zfox23, line 5: what is a "WSS URL" ?

RebeccaStankus: It's the url that looks like this: `wss://loadbalancer-api-staging-04.highfidelity.io:8001/` I don't know what to call it.

zfox23 (approved):

PR #25, opened 2021-03-10
Add feature detection for browser compatibility

luiscuenca:
This PR increases compatibility by setting the ts compiler target to `ES5`. A check for required features will be performed before executing the API code. This way the browser will prompt the user via console before other errors occur.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,6 +5,8 @@ import { checkBrowserCompatibility } from "./utilities/HiFiUtilities";
 let isBrowserContext = typeof self !== 'undefined';
 if (isBrowserContext) {
     exports.HiFiAPIVersion = HIFI_API_VERSION;
+    exports.apiVersion = HIFI_API_VERSION;
+    exports.checkBrowserCompatibility = checkBrowserCompatibility;
     checkBrowserCompatibility();
 }
 
@@ -55,10 +57,6 @@ exports.HiFiAxisConfiguration = HiFiAxisConfiguration;
 // Short synonyms for the above start here!
 // Please let us know if any of these `exports` cause namespace collisions
 // in your application.
-
-if (isBrowserContext) {
-    exports.apiVersion = HIFI_API_VERSION;
-}
 
 exports.Communicator = HiFiCommunicator;
 exports.ConnectionStates = HiFiConnectionStates;
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,7 +2,11 @@ declare var HIFI_API_VERSION: string;
 
 // Check for browser compatibility
 import { checkBrowserCompatibility } from "./utilities/HiFiUtilities";
-checkBrowserCompatibility();
+let isBrowserContext = typeof self !== 'undefined';
+if (isBrowserContext) {
+    exports.HiFiAPIVersion = HIFI_API_VERSION;
+    checkBrowserCompatibility();
+}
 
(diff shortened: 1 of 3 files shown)


src/utilities/HiFiUtilities.ts
```diff
@@ -186,3 +186,40 @@ export function preciseInterval(callback: Function, intervalMS: number): any {
...
+        "window.RTCPeerConnectionIceEvent",
+        "window.RTCRtpReceiver",
+        "window.RTCRtpSender",
+        "window.RTCRtpTransceiver",
+        "window.RTCSctpTransport",
+        "window.RTCSessionDescription"
+    ]
+    for (let i = 0; i < requiredFeatures.length; i++) {
+        if (typeof(eval(requiredFeatures[i])) === "undefined") {
+            console.log("HiFi Audio API: The browser does not support: " + requiredFeatures[i]);
```
zfox23, line 216:
```suggestion
            HiFiLogger.error("HiFi Audio API: The browser does not support: " + requiredFeatures[i]);
```

src/utilities/HiFiUtilities.ts
```diff
@@ -186,3 +186,40 @@ export function preciseInterval(callback: Function, intervalMS: number): any {
...
+            if (requiredFeatures[i] === "navigator.mediaDevices.getUserMedia") {
+                console.log("HiFi Audio API: It might be preventing access to this feature on insecure contexts.")
```
zfox23, line 218: Please replace with `HiFiLogger.error()`, i.e.:
```suggestion
                HiFiLogger.error("HiFi Audio API: It might be preventing access to this feature on insecure contexts.")
```

zfox23: Also, perhaps consider this language:
```suggestion
                HiFiLogger.error("HiFi Audio API: Your browser may be preventing access to this feature if you are running in an insecure context, i.e. an `http` server.")
```

src/utilities/HiFiUtilities.ts
```diff
@@ -186,3 +186,40 @@ export function preciseInterval(callback: Function, intervalMS: number): any {
...
+            }
+            return false;
+        }
+    }
+    console.log("HiFi Audio API: The browser supports all feature requirements.")
```
zfox23, line 223: I don't think we should bother logging anything in this case!

src/index.ts
```diff
@@ -1,5 +1,9 @@
 declare var HIFI_API_VERSION: string;
 
+// Check for browser compatibility
+import { checkBrowserCompatibility } from "./utilities/HiFiUtilities";
+checkBrowserCompatibility();
```
zfox23, line 5: please make sure this only runs in the browser context, not in NodeJS!

zfox23: Also, please add an `exports.checkBrowserCompatibility = checkBrowserCompatibility;` for external users :)

src/utilities/HiFiUtilities.ts
luiscuenca: Fixed

src/utilities/HiFiUtilities.ts
luiscuenca: Yeah that reads better. Changed

src/utilities/HiFiUtilities.ts
luiscuenca: Sounds good

src/index.ts
luiscuenca: Added

zfox23 (dismissed):
i'm curious about the downsides of targeting ES5, but we'll talk about that separately. approved

zfox23 (approved):

PR #36, opened 2021-03-18
Search accessibility

RebeccaStankus:
https://highfidelity.atlassian.net/browse/HIFI-365


typedoc-theme/assets/css/hifi.css
```diff
@@ -34,3 +39,37 @@
 body {
     font-family: 'Graphik', "Segoe UI", sans-serif;
 }
+
+@media screen and (min-width: 600px) {
```
zfox23, line 43: why is this CSS rule necessary? it results in the search bar being quite small on many desktop screens, which looks strange. if you didn't have this rule, what would break?

image

zfox23: this is what the same screen looks like without this rule:

image

RebeccaStankus: We wanted the search bar to be smaller. This was intentional from our discussion the other day.

zfox23 (approved):

PR #42, opened 2021-03-22
Allow user to set jitter buffer parameters via HiFiCommunicator constructor; Do some refactoring for clarity and futureproofing

zfox23:
https://highfidelity.atlassian.net/browse/HIFI-445

### DIFF
--- a/src/classes/HiFiCommunicator.ts
+++ b/src/classes/HiFiCommunicator.ts
@@ -102,7 +102,8 @@ export class HiFiCommunicator {
      * @param transmitRateLimitTimeoutMS - User Data updates will not be sent to the server any more frequently than this number in milliseconds.
      * @param userDataStreamingScope - Cannot be set later. See {@link HiFiUserDataStreamingScopes}.
      * @param hiFiAxisConfiguration - Cannot be set later. The 3D axis configuration. See {@link ourHiFiAxisConfiguration} for defaults.
-     * @param raviSessionParams - Cannot be set later. The RAVI session parameters used when connecting to the API servers. See {@link RaviSessionParams} for details.
+     * @param webrtcSessionParams - Cannot be set later. Extra parameters used for configuring the underlying WebRTC connection to the API servers.
+     * These settings are not frequently used; they are primarily for specific jitter buffer configurations. See {@link RaviSessionParams} for details.
      */
     constructor({
         initialHiFiAudioAPIData = new HiFiAudioAPIData(),
@@ -111,7 +112,7 @@ export class HiFiCommunicator {
         transmitRateLimitTimeoutMS = HiFiConstants.DEFAULT_TRANSMIT_RATE_LIMIT_TIMEOUT_MS,
         userDataStreamingScope = HiFiUserDataStreamingScopes.All,
         hiFiAxisConfiguration,
-        raviSessionParams
+        webrtcSessionParams
     }: {
         initialHiFiAudioAPIData?: HiFiAudioAPIData,
         onConnectionStateChanged?: Function,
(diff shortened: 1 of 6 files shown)


src/classes/HiFiCommunicator.ts
```diff
@@ -98,21 +102,24 @@ export class HiFiCommunicator {
      * @param transmitRateLimitTimeoutMS - User Data updates will not be sent to the server any more frequently than this number in milliseconds.
      * @param userDataStreamingScope - Cannot be set later. See {@link HiFiUserDataStreamingScopes}.
      * @param hiFiAxisConfiguration - Cannot be set later. The 3D axis configuration. See {@link ourHiFiAxisConfiguration} for defaults.
+     * @param raviSessionParams - Cannot be set later. The RAVI session parameters used when connecting to the API servers. See {@link RaviSessionParams} for details.
```
hd23408, line 105: I don't want to confuse customers with the "RAVI" name more than we have to. Let's call this something like "webrtcSessionParams" instead, and make the comment something like "Extra parameters used for configuring the underlying WebRTC connection to the API servers. These settings are not frequently used; they are primarily for specific jitter buffer configurations." (And then I'm OK with continuing to link to the RaviSessionParams docs.)

src/libravi/RaviSession.ts
```diff
@@ -22,6 +22,25 @@ export enum RaviSessionStates {
   CLOSED = "closed"
 };
 
+export interface RaviSessionParams {
+  /**
+   * The minimum jitter buffer duration.
+   * 
+   * The lower the jitter buffer duration, the lower the round-trip audio latency.
```
hd23408, line 29: For the minimum jitter buffer duration, I'm concerned that the comments about "setting it too low" will confuse people. Let's change this comment to something like: 

"The minimum jitter buffer duration (in seconds). In practice, this should always be set to 0 (the default). Increasing the minimum jitter buffer means that all audio will always be buffered at least by this amount of time, which is rarely desirable. You may, however, want to set the maximum jitter buffer duration if you're increasing frequent audio drop-outs; refer to audioMaxJitterBufferDuration for more details."

src/libravi/RaviSession.ts
```diff
@@ -22,6 +22,25 @@ export enum RaviSessionStates {
...
+   * However: set the jitter buffer duration too low on an unreliable connection, and users receiving an audio stream may hear interruptions or
+   * "drop-outs" in the audio stream. Set the jitter buffer duration high to reduce the possibility of dropouts at the cost of higher round-trip audio latency.
+   */
+  audioMinJitterBufferDuration?: number;
+  /**
+   * The maximum jitter buffer duration.
```
hd23408, line 35: add "(in seconds). The default is 1 second."

src/libravi/RaviSession.ts
```diff
@@ -22,6 +22,25 @@ export enum RaviSessionStates {
...
+   * 
+   * The lower the jitter buffer duration, the lower the round-trip audio latency.
+   * However: set the jitter buffer duration too low on an unreliable connection, and users receiving an audio stream may hear interruptions or
+   * "drop-outs" in the audio stream. Set the jitter buffer duration high to reduce the possibility of dropouts at the cost of higher round-trip audio latency.
```
hd23408, line 39: "Set the jitter buffer duration high to reduce the possibility of dropouts at the cost of potentially higher round-trip audio latency on poor connections."

hd23408 (changes requested):
Looks good, just some comment comments!

hd23408 (approved):

PR #60, opened 2021-04-02
Github Actions run Jest unit and integration tests on every push

RebeccaStankus:


.github/workflows/run-unit-tests.yml
```diff
@@ -1,37 +1,68 @@
...
       - main
+      - gha
+ 
 
 jobs:
-  Run-Jest-Unit-Tests:
+  Run-Jest-Tests:
     runs-on: ubuntu-latest
-    if: github.event.pusher.name != 'hifibuild'
+    if: github.event.pusher.namen != 'hifibuild'
```
luiscuenca, line 22: Is this a typo?

RebeccaStankus: oh, yeah...thanks!

luiscuenca (dismissed):

luiscuenca (approved):

joy-highfidelity (approved):
✨

hd23408 (commented):
This is exciting! Thank you @RebeccaStankus !

I'd like to have us NOT run the admin REST API tests on every push to hifi-spatial-audio-js, because there isn't anything that could be changed in the hifi-spatial-audio-js code that would affect the behavior of REST API calls.

So, I think that tests/integration/rest.integration.test.ts already separates out all the REST API tests, correct? If so, what I'd like to happen is for us NOT to run the REST tests as part of the GHA runs, and instead just run the HiFiCommunicator.integration.test.ts tests. Thanks!

RebeccaStankus:
@hd23408 I went ahead and reintroduced 'smoke' tests for the category we run on deploy and put REST tests in there.

luiscuenca (approved):

PR #61, opened 2021-04-05
Add hifiCommunicator.adjustPersonalVolume with documentation

sabrina-shanman:


src/classes/HiFiCommunicator.ts
```diff
@@ -257,6 +257,54 @@ export class HiFiCommunicator {
...
     }
 
+    /**
+     * Adjusts the perceived volume of another user for this communicator's current connection only.
+     * This can be used to provide a more comfortable listening experience for the client. If you need to perform moderation actions which apply to all users, use the {@link https://docs.highfidelity.com/rest/latest/index.html|Administrative REST API}.
+     * 
+     * To use this command, the communicator must currently be connected to a space. You can connect to a space using {@link connectToHiFiAudioAPIServer}.
+     * 
+     * @param hashedVisitId  The hashed visit ID of the user whose volume will be adjusted.
+     * Use {@link addUserDataSubscription} and {@link HiFiCommunicator.onUsersDisconnected} to keep track of the hashed visit IDs of currently connected users.
```
hd23408, line 267: Point developers to the `ReceivedHiFiAudioAPIData` object as well, and specifically the fact that they can find the appropriate `hashedVisitId` for a given `providedUserID` (if they're specifying `provideUserIDs`) by examining the `ReceivedHiFiAudioAPIData` objects returned from the data subscription. 

Side note: I don't love that from the customer/developer side, but I'm struggling to think of a better way to do this. Ideally we would allow the developer to use `providedUserID` here instead of `hashedVisitId`, but since we don't require `providedUserID` (nor do we enforce uniqueness of it) I don't see an obvious way to get there. So, for now, I think using `hashedVisitId` and just explaining in the comments here how a developer could map from `providedUserID` to `hashedVisitId` is gonna be the best way to go.

src/classes/HiFiCommunicator.ts
```diff
@@ -257,6 +257,54 @@ export class HiFiCommunicator {
...
+     *
+     * @param gain  The relative volume to apply to the other user. By default, this is `1.0`. The gain can be any value greater or equal to `0.0`.
+     * For example: a gain of `2.0` will double the volume of the user, while a gain of `0.5` will halve the user's volume. A gain of `0.0` will effectively mute the user.
+     * 
+     * @returns If this operation is successful, the Promise will resolve with `{ success: true, personalVolumeAdjustResponse: <The response to `audionet.personal_volume_adjust` from the server in Object format> }`.
+     * If unsuccessful, the Promise will reject with `{ success: false, error: <an error message> }`.
+     */
+    async adjustPersonalVolume(visitIdHash: String, gain: Number): Promise<any> {
```
hd23408, line 275: I'm a little concerned that this method name will imply that you're adjusting your OWN volume, rather than someone else's volume. (Also we should probably explicitly say 'gain' rather than 'volume'.) Thoughts for alternatives? Maybe just `adjustOtherUserGain`?

hd23408 (changes requested):
Looks great! Only comments are around documentation and naming. Thank you Sabrina!! ❤️

src/classes/HiFiCommunicator.ts
sabrina-shanman: I agree that using the term `gain` instead would be better for the sake of more consistent terminology.

I also see the potential for confusion with the use of the word, "personal." I would like to convey somehow that the gain change of the other user only applies to this particular client's stream only. Something like, "adjust another user's gain, but just for me." I'll need to think about it a little bit.

sabrina-shanman: Thoughts on these? They are a bit verbose:

- `setOtherUserGainForMyStream`
- `setOtherUserGainLocally`

hd23408: `setOtherUserGainForThisConnection` maybe? Naming things is HARD.

src/classes/HiFiMixerSession.ts
```diff
@@ -198,6 +199,60 @@ export class HiFiMixerSession {
         });
     }
 
+    /**
+     * Sends the command `audionet.personal_volume_adjust` to the mixer. This adjusts the volume of another user session, only from the perspective of this client, for the current connection only.
+     * 
+     * @returns If this operation is successful, the Promise will resolve with `{ success: true, audionetAdjustPersonalVolumeResponse: <The response to `audionet.personal_volume_adjust` from the server in Object format> }`.
+     * If unsuccessful, the Promise will reject with `{ success: false, error: <an error message> }`.
+     */
+    async adjustPersonalVolume(visitIdHash: String, gain: Number): Promise<any> {
```
zfox23, line 208: Instead of returning `Promise<any>` here, please create a new `interface` corresponding to the structure of the data returned, then return `Promise<[that interface]>`.

For example, you may create a new interface at the top of the file like:
```ts
export interface AdjustPersonalVolumeResponse {
    success: boolean,
    error?: string,
    audionetAdjustPersonalVolumeResponse?: AudionetAdjustPersonalVolumeResponse
}
```

you'll then need another interface like:

```ts
interface AudionetAdjustPersonalVolumeResponse {
    <the data returned from the server>
}
```

If you don't know how `interface AudionetAdjustPersonalVolumeResponse` would be structured, first ask yourself if there's a good reason why that might be. if you're _sure_ there's a good reason, then you don't need that second interface, and you can replace `audionetAdjustPersonalVolumeResponse?: AudionetAdjustPersonalVolumeResponse` in the first `interface` with `audionetAdjustPersonalVolumeResponse?: any`.

in general, it's good practice to use `interface`s instead of `any` whenever possible. anytime there's an `any`, there's an increased chance of error.

zfox23 (changes requested):
wow! overall, i'm really impressed - this is super thorough. one change requested, please! and i look forward to the new name of the command - i prefer verbosity, so something like `setOtherUserOutputVolumeForMyConnection` would be great, IMO :)

zfox23 (approved):

hd23408 (approved):

PR #64, opened 2021-04-06
Update server connection tests to use hostname

RebeccaStankus:


RebeccaStankus:
Closing as this will be covered when pr 60 is merged and is NOT necessary for tests to pass now.

PR #69, opened 2021-04-10
WIP: Implement / explore the axis configuration feature

samcake:
This pr is a wip demonstrating the axis configuration feature 

it is NOT complete, 
specifically all the possible coordinate system conventions are not implemented.
I just implemented all the varieties of Y up first (right handed and left handed)

the main missing feature is with the euler angles represenation conversion.

Happy to meet and talk about it when you have time for this,
Over

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,9 +6,10 @@
   "packages": {
     "": {
       "name": "hifi-spatial-audio",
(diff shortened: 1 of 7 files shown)


luiscuenca:
We will make this feature available as a helper function

PR #84, opened 2021-04-22
Make setOtherUserGainForThisConnection use input channel to enable real-time updates

sabrina-shanman:


src/classes/HiFiAudioAPIData.ts
```diff
@@ -351,6 +351,12 @@ export function eulerFromQuaternion(quat: OrientationQuat3D, order: OrientationE
     return new OrientationEuler3D({ pitchDegrees: RAD_TO_DEG * pitch, yawDegrees: RAD_TO_DEG * yaw, rollDegrees: RAD_TO_DEG * roll });
 }
 
+/**
+ * Instantiations of this class define a map between hashed visit IDs and the gains of other users as percieved by the current user's session. This class is used by {@link HiFiAudioAPIData.otherUserGainQueue} to keep track of which gain changes need to be sent to the server.
+ * See also {@link HiFiCommunicator.setOtherUserGainForThisConnection}, which allows you to set the desired gain for a user and also send the data to the server. If you need to perform moderation actions which apply to all users, use the {@link https://docs.highfidelity.com/rest/latest/index.html|Administrative REST API}.
+ */
+export type OtherUserGainMap = { [key: string]: number };
+
```
hd23408, line 359: This map is only used internally and isn't meant to be adjusted or inspected directly by the user (i.e. the developer), correct? Let's call that out explicitly in these comments. Is it possible to make it private and/or otherwise hidden from documentation and developers?

src/classes/HiFiAudioAPIData.ts
```diff
@@ -445,15 +451,26 @@ export class HiFiAudioAPIData {
...
+     * This can be used to provide a more comfortable listening experience for the client. However, setting a gain here does not on its own change the gain of another user. See instead {@link HiFiCommunicator.setOtherUserGainForThisConnection}, which allows you to set the desired gain for a user and also send the data to the server. If you need to perform moderation actions which apply to all users, use the {@link https://docs.highfidelity.com/rest/latest/index.html|Administrative REST API}.
+     * 
+     * Each key is the hashed visit ID of a user whose gain will be affected, and its associated value is the desired gain of the other user. Any gain set here only affects the current user's session. Other users beside the current user will not perceive a difference in volume when a gain setting is set here.
+     * 
+     * These gains are not persistent. If the other user's session restarts, then their gain will be reset for the current user. If the current user's session restarts, then all other user gains will be reset.
+     */
+    otherUserGainQueue: OtherUserGainMap;
```
hd23408, line 463: See above. Is there a use case for allowing developers to interact with this directly? I think it might make more sense to hide this from devs (and therefore from the constructor parameters) and just keep it private and initialize it in the ctor. Developers should (I think?) _just_ use the `setOtherUserGainForThisConnection` method rather than trying to understand what's going on with the gain queue.

hd23408 (changes requested):

src/classes/HiFiAudioAPIData.ts
sabrina-shanman: I can convert the comment to something that won't show up in the documentation, and emphasize that it is internal.

src/classes/HiFiAudioAPIData.ts
sabrina-shanman: You're right, I don't think there's a use case for using this directly. Removing the ability to set it in the constructor seems reasonable. But we can't make it private because it's used by `HiFiMixerSession` to sync the other user gains. One alternative option is to have a separate class called `InternalHiFiAudioAPIData` which stores data like this. But for now, keeping this data all in the same place seems better.

src/classes/HiFiAudioAPIData.ts
hd23408: That sounds excellent, thank you!

src/classes/HiFiAudioAPIData.ts
hd23408: Yep, that's fine for now, thank you! ❤️

src/classes/HiFiCommunicator.ts
```diff
@@ -276,40 +276,16 @@ export class HiFiCommunicator {
...
-                errorReason = JSON.stringify(errorSettingUserGain);
-            }
-            let errMsg = `Error when setting other user gain for this connection. Error: \n${errorReason}`;
-            return Promise.reject({
-                success: false,
-                error: errMsg,
-                audionetSetOtherUserGainForThisConnectionResponse: errorSettingUserGain.audionetSetOtherUserGainForThisConnectionResponse
-            });
-        }
+    async setOtherUserGainsForThisConnection(otherUserGainMap: OtherUserGainMap): Promise<SetOtherUserGainsForThisConnectionResponse> {
```
hd23408, line 282: @sabrina-shanman Sorry, one final thing -- because people already have `setOtherUserGainForThisConnection` in production, can you please add a wrapper function of `async setOtherUserGainForThisConnection(visitIdHash: String, gain: Number): Promise<SetOtherUserGainForThisConnectionResponse>` to this for backwards compatibility and the "single user" use case?

In other words, there should continue to be a `setOtherUserGainForThisConnection` method, but internally it should just create the single-user OtherUserGainMap and hand it into this one.

sabrina-shanman: Oh, right! I'll add that back.

hd23408: Similarly...if it's possible, it would be great to add an alias of `SetOtherUserGainForThisConnectionResponse` for the newly renamed `SetOtherUserGainsForThisConnectionResponse`. If that's not possible, I think my preference would be to leave it singular (for backwards compatibility) despite the naming difference with the primary (plural) method.

PR #85, opened 2021-04-23
Working tests on dev stacks

RebeccaStankus:
Fixes:
* hyphenated non-preset stacknames can be parsed correctly
* tests actually grab stack data from the auth file for non-preset stacks (it was just exiting if the stackname didn't match a preset before)
* stackname input includes 'highfidelity.com' or '.highfidelity.io' so both extensions can be used

### DIFF
--- a/tests/smoke/rest.smoke.test.ts
+++ b/tests/smoke/rest.smoke.test.ts
@@ -6,30 +6,29 @@ import { TestUser } from '../testUtilities/TestUser';
 import { HiFiConnectionStates } from "../../src/classes/HiFiCommunicator";
 
 let args = require('minimist')(process.argv.slice(2));
-let stackname = args.stackname || process.env.hostname || "api-staging-latest";
+let stackname = args.stackname || process.env.hostname || "api-staging-latest.highfidelity.com";
 console.log("_______________STACKNAME_______________________", stackname);
-let stackURL = `https://${stackname}.highfidelity.com`;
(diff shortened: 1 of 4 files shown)


tests/smoke/rest.smoke.test.ts
```diff
@@ -126,7 +125,7 @@ describe('HiFi API REST Calls', () => {
             expect(settingsJSON['global-frequency-rolloff']).toBe(null);
             expect(settingsJSON['global-attenuation']).toBe(null);
             expect(settingsJSON['client-limit']).toBe(null);
-            expect(settingsJSON['max-client-limit']).toBe(20);
+            expect(settingsJSON['max-client-limit']).toBeLessThanOrEqual(200);
```
hd23408, line 128: FYI, this will be changing with the work Seth is doing in https://github.com/highfidelity/speakeasy-infra/pull/366 -- in retrospect it might be best just to pull it out if you want? Up to you and @bridie-hifi .

hd23408 (dismissed):

tests/smoke/rest.smoke.test.ts
RebeccaStankus: oh, yeah, I meant to not include that in this PR anyway. I'll get rid of it for now butt leave a comment in case we want to add some sort of check back again after Seth's work.

hd23408 (approved):

PR #86, opened 2021-04-26
Fix HIFI-579 - except this isn't the best fix

zfox23:
https://highfidelity.atlassian.net/browse/HIFI-579


hd23408:
The better fix was done with v1.34.0 of the mixer and released on May 5/6 2021.

PR #91, opened 2021-04-28
Add mute callback and handle mute instruction from server

sabrina-shanman:


src/classes/HiFiMixerSession.ts
```diff
@@ -583,9 +682,52 @@ export class HiFiMixerSession {
...
+        // - When the connection ends, the client is allowed to unmute, which for now is equivalent to an admin unmute
+        this._adminPreventsInputAudioUnmuting = mutedByAdmin;
+        return await this._setMuted(mutedByAdmin || this._lastSuccessfulInputAudioMutedValue, false, muteReason);
+    }
+
+    async _setMuted(newMutedValue: boolean, tryToStopMicStream: boolean, muteReason: MuteReason): Promise<boolean> {
+        let success = true;
+        if (muteReason == MuteReason.CLIENT) {
+            if (this._adminPreventsInputAudioUnmuting && !newMutedValue) {
+                HiFiLogger.warn(`Couldn't set mute state: Muted by admin.`);
```
hd23408, line 707: Is this also managed on the mixer side? For instance, if User A gets muted by an admin and tries to unmute themself, the `_adminPreventsInputAudioUnmuting` flag prevents the unmuting from happening here. However, if User A is a l33t hax0r, and manages to change `_adminPreventsInputAudioUnmuting` to `false`, could they then unmute themselves? Or does the mixer say "Nope, not unmuting, admin said so."

sabrina-shanman: _adminPreventsInputAudioUnmuting is set to `true` when the client is muted on the mixer side, i.e. even if they were l33t hax0r and set it to `false` on the javascript side, it would do them no good as the mixer would continue ignoring their audio.

hd23408 (approved):

PR #101, opened 2021-05-09
Fine-tuning of connection behavior

hd23408:
**HIFI-603**: HiFiCommunicator's `connectToHiFiAudioAPIServer` method will now return immediately if the connection is already, uh, connected. (In the end I opted to make this a "success" rather than a failure because I figure it did accomplish the presumed goal, which was to connect - but I waffled a bit, especially because it has a different return (a message rather than results of audionet.init.)

**HIFI-602**: New `getConnectionState()` method on HiFiCommunicator so that users can easily check the status of the connection at any time.

**HIFI-586**: Now appropriately `await`s disconnect attempts. This was resulting in issues when trying to immediately retry connection attempts after they had failed.

**Unrelated changes:**
- Adjusted a comment slightly 
- Changed all of the "\n"s that were embedded in template literals to be actual newlines, because I was getting sick of seeing "\n"s scattered all throughout the text of log messages. This may be a controversial change, because it breaks indentation (so that we don't just end up replacing "\n" with lots of spaces). All I can say is that if we want to use template literals, we have to take the good with the bad (unless someone can tell me how to get the console output to output an actual newline instead of a \n when it's embedded inside a template literal, in which case I will happily switch back to them).

### DIFF
--- a/src/classes/HiFiCommunicator.ts
+++ b/src/classes/HiFiCommunicator.ts
@@ -234,8 +234,8 @@ export class HiFiCommunicator {
 
         if (this._mixerSession.getCurrentHiFiConnectionState() === HiFiConnectionStates.Connected) {
             let errMsg = `Session is already connected! If you need to reset the connection, please disconnect fully and call this method again.`;
-            return Promise.reject({
-                success: false,
+            return Promise.resolve({
+                success: true,
                 error: errMsg
             });
         }
--- a/src/classes/HiFiCommunicator.ts
+++ b/src/classes/HiFiCommunicator.ts
@@ -348,6 +348,19 @@ export class HiFiCommunicator {
     }
 
     /**
+     * @returns The current state of the connection to High Fidelity, as one of the HiFiConnectionStates.
+     * This will return null if the current state is not available (e.g. if the HiFiCommunicator
+     * is still in the process of initializing its underlying HiFiMixerSession).
+     */
+    getConnectionState(): HiFiConnectionStates {
+        if (this._mixerSession) {
+            return this._mixerSession.getCurrentHiFiConnectionState();
+        } else {
+            return null;
+        }
+    }
+
+    /**
      * Use this function to set the `MediaStream` associated with the user. This `MediaStream` will be sent up to the High Fidelity Audio Servers and
      * mixed with other users' audio streams. The resultant mixed stream will be sent to all connected clients.
      *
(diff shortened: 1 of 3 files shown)


src/libravi/RaviSignalingConnection.ts
```diff
@@ -370,6 +370,14 @@ class RaviSignalingWebSocketImplementation {
...
    */
   _open(socketAddress: string) {
+
+    // If we already have an open websocket, just log and return
+    if (this._webSocket && this._webSocket.readyState === crossPlatformWebSocket.OPEN) {
+        RaviUtils.err("There is already an open WebSocket connection. To reconnect, first close the existing WebSocket and then attempt to open again.",
+                "RaviSignalingWebSocketImplementation");
+        return;
+    }
+
```
hd23408, line 380: (Note, this change is also captured in https://github.com/highfidelity/libravi/pull/445/files on the RAVI side.)

src/classes/HiFiCommunicator.ts
```diff
@@ -225,13 +225,21 @@ export class HiFiCommunicator {
...
-            let errMsg = `\`this._mixerSession\` is falsey!`;
+            let errMsg = `\`this._mixerSession\` is falsey; try creating a new HiFiCommunicator and starting over.`;
             return Promise.reject({
                 success: false,
                 error: errMsg
             });
         }
 
+        if (this._mixerSession.getCurrentHiFiConnectionState() === HiFiConnectionStates.Connected) {
+            let msg = `Session is already connected! If you need to reset the connection, please disconnect fully and call this method again.`;
```
zfox23, line 236:
```suggestion
            let msg = `Session is already connected! If you need to reset the connection, please disconnect fully with `disconnectFromHiFiAudioAPIServer()` and call this method again.`;
```

src/classes/HiFiCommunicator.ts
```diff
@@ -248,11 +256,13 @@ export class HiFiCommunicator {
             let webRTCSignalingAddress = `wss://${signalingHostURLSafe}:${signalingPort}/?token=`;
             this._mixerSession.webRTCAddress = `${webRTCSignalingAddress}${hifiAuthJWT}`;
 
-            HiFiLogger.log(`Using WebRTC Signaling Address:\n${webRTCSignalingAddress}<token redacted>`);
+            HiFiLogger.log(`Using WebRTC Signaling Address:
```
zfox23, line 259: was this change on purpose? if so, what is it for?

hd23408: Yep, this is because the "\n"s that were embedded in the template literals (the backticked strings) made my console logs full of explicit "\n"s instead of ACTUAL linebreaks, and it has been making me crazy and it's really hard for me to debug. If there's a way to get the console output to output an _actual_ newline instead of a "\n" when "\n" is embedded inside a template literal, I'm happy to use that instead and go back to "\n"s. Otherwise, if we're using template literals, we should probably be using literal linebreaks rather than "\n"s. (Or we could stop using template literals, which I'm also OK with because TBH I don't like them -- backtick is a very special character for bash and others, and I'd rather use real quotes for JavaScript strings. :-) But I understand that JS programmers like 'em!)

zfox23 (approved):
thanks :)

luiscuenca (approved):

PR #102, opened 2021-05-12
Use Puppeteer to run tests in browser env

RebeccaStankus:
https://highfidelity.atlassian.net/browse/HIFI-638


RebeccaStankus:
This may not be necessary now after https://github.com/highfidelity/hifi-spatial-audio-js/pull/103

PR #107, opened 2021-05-17
Fix spatial audio JS client not responding to admin mute command

sabrina-shanman:
https://highfidelity.atlassian.net/browse/HIFI-642

### DIFF
--- a/src/classes/HiFiMixerSession.ts
+++ b/src/classes/HiFiMixerSession.ts
@@ -485,9 +485,9 @@ export class HiFiMixerSession {
                 if (instructionName === "mute") {
                     let shouldBeMuted: boolean;
                     if (instructionArguments.length >= 1) {
-                        if (instructionArguments[0] === 1) {
+                        if (instructionArguments[0] === true) {
                             shouldBeMuted = true;
-                        } else if (instructionArguments[0] === 0) {
+                        } else if (instructionArguments[0] === false) {
                             shouldBeMuted = false;
                         }
                     }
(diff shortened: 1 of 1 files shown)


hd23408 (approved):

src/classes/HiFiMixerSession.ts
```diff
@@ -485,9 +485,9 @@ export class HiFiMixerSession {
                 if (instructionName === "mute") {
                     let shouldBeMuted: boolean;
                     if (instructionArguments.length >= 1) {
-                        if (instructionArguments[0] === 1) {
+                        if (instructionArguments[0] === true) {
                             shouldBeMuted = true;
-                        } else if (instructionArguments[0] === 0) {
+                        } else if (instructionArguments[0] === false) {
```
hd23408, line 490: Could this perhaps instead be
`shouldBeMuted = instructionArguments[0]`?

hd23408 (changes requested):

src/classes/HiFiMixerSession.ts
hd23408: (Or do we not trust the incoming value of instructionArguments[0]?)

RebeccaStankus (dismissed):

src/classes/HiFiMixerSession.ts
sabrina-shanman: We probably do trust the value of `instructionArguments[0]`, but for completeness (and Typescript type correctness), I've simplified the code but still check if it's a boolean.

hd23408 (approved):

PR #112, opened 2021-05-19
Updated attenuation tests

RebeccaStankus:
https://highfidelity.atlassian.net/browse/HIFI-645

### DIFF
--- a/tests/smoke/rest.smoke.test.ts
+++ b/tests/smoke/rest.smoke.test.ts
@@ -569,7 +569,7 @@ describe('HiFi API REST Calls', () => {
                 zone1Data['id'] = responseJSON[1].id;
                 zone2Data['id'] = responseJSON[0].id;
             }
-            expect(responseJSON.map((a: { id: any; }) => a.id).sort()).toEqual([zone1Data, zone2Data].map(a => a.id).sort());
+            expect(responseJSON.map((a: { id: number; }) => a.id).sort()).toEqual([zone1Data, zone2Data].map(a => a.id).sort());
 
             // Create one zone via space `settings/zones/create` POST request
             returnMessage = await fetch(`${stackURL}/api/v1/spaces/${spaceID}/settings/zones?token=${adminToken}`, {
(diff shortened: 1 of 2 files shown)


tests/testUtilities/testUtils.ts
```diff
@@ -140,6 +140,7 @@ export interface ZoneData {
 
 export interface AttenuationData {
```
sabrina-shanman, line 141: So, does the returned data usually include "id", or is it "zone-attenuation-id"?

tests/smoke/rest.smoke.test.ts
```diff
@@ -656,24 +656,28 @@ describe('HiFi API REST Calls', () => {
...
                 "listener-zone-id": zone1Data.id,
                 "source-zone-id": zone3Data.id,
                 "za-offset": -5
             };
             attenuation4Data = {
                 "attenuation": 0.5,
+                "frequency-rolloff": 12,
                 "listener-zone-id": zone1Data.id,
                 "source-zone-id": zone4Data.id,
                 "za-offset": -5
```
sabrina-shanman, line 683: We should try testing `null` as a value for one of the frequency rolloffs and one of the attenuations, now that that's a feature.

sabrina-shanman (changes requested):

tests/smoke/rest.smoke.test.ts
```diff
@@ -708,7 +712,7 @@ describe('HiFi API REST Calls', () => {
             expect(responseJSON).toEqual([attenuation3Data]);
```
sabrina-shanman, line 712: Hmm, if this is working, then I think the new documentation is wrong. It currently says the ID key is "zone-attenuation-id". I suspect this is a regression that I introduced.

tests/smoke/rest.smoke.test.ts
RebeccaStankus: Actually, when I tried inputting null, the fetch request failed.

sabrina-shanman (approved):

PR #114, opened 2021-05-25
WIP: simpler util for configuring world-frame and transforming between world- and hifi-frame

AndrewMeadows:
This PR introduces a simpler Util for converting 3D points between world- and hifi-frame.

The intention was to address [HIFI-577](https://highfidelity.atlassian.net/browse/HIFI-557), however so far this PR only fixes errors in frame transforms and does not yet remove the transform logic from the HiFi Spatial Audio API into a distinct helper utility.


src/classes/HiFiAudioAPIData.ts
```diff
@@ -32,6 +32,188 @@ export class Point3D {
...
+
+    /**
+      * Negates this Point3D.
+      */
+    negate() {
+        this.x = -this.x;
+        this.y = -this.y;
+        this.z = -this.z;
+    }
+}
```
AndrewMeadows, line 58: The **Point3D** `dot`, `cross`, and `negate` methods above were necessary to do the math below.

src/classes/HiFiAxisConfiguration.ts
```diff
@@ -4,270 +4,63 @@
...
-
-        return isValid;
-    }
-
-    /** 
-     * ⚠ WARNING ⚠ The code in this function IS wrong.
-     * TODO: implement the function, just a NO OP at the moment.
-     * 
-     * @param axisConfiguration 
-     * @param inputPoint3D
```
AndrewMeadows, line 180: The **WorldFrameConfiguration** replaces **HiFiAxisConfiguration** and is simpler. There are only three degrees of freedom when choosing a sensible coordinate frame:

**(1)** forward direction
**(2)** up direction
**(3)** left- or right-handedness

The third "direction" of can be derived.

src/classes/HiFiMixerSession.ts
```diff
@@ -203,14 +215,15 @@ export class HiFiMixerSession {
...
-    constructor({ userDataStreamingScope = HiFiUserDataStreamingScopes.All, onUserDataUpdated, onUsersDisconnected, onConnectionStateChanged, onMuteChanged }: { userDataStreamingScope?: HiFiUserDataStreamingScopes, onUserDataUpdated?: Function, onUsersDisconnected?: Function, onConnectionStateChanged?: Function, onMuteChanged?: OnMuteChangedCallback }) {
+    constructor({ userDataStreamingScope = HiFiUserDataStreamingScopes.All, onUserDataUpdated, onUsersDisconnected, onConnectionStateChanged, onMuteChanged, coordFrameUtil }: { userDataStreamingScope?: HiFiUserDataStreamingScopes, onUserDataUpdated?: Function, onUsersDisconnected?: Function, onConnectionStateChanged?: Function, onMuteChanged?: OnMuteChangedCallback, coordFrameUtil?: HiFiCoordinateFrameUtil}) {
         this.webRTCAddress = undefined;
         this.userDataStreamingScope = userDataStreamingScope;
         this.onUserDataUpdated = onUserDataUpdated;
         this.onUsersDisconnected = onUsersDisconnected;
         this._mixerPeerKeyToStateCacheDict = {};
         this._lastSuccessfulInputAudioMutedValue = false;
         this.onMuteChanged = onMuteChanged;
+        this._coordFrameUtil = coordFrameUtil;
```
AndrewMeadows, line 226: `this._coordFrameUtil` is allowed to be null.  We only transform between coordinate frames when it is non-null.

src/classes/HiFiAudioAPIData.ts
```diff
@@ -32,6 +32,188 @@ export class Point3D {
...
+        //
+        // | <-forward-> |
+        // | <-  up   -> |
+        // | <- right -> |
+        //
+        let right = Point3D.cross(forward, up);
+        if (isLeft) {
+            // Note: for left-handed systems we reflect 'right' to actually be 'left'
+            right.negate();
+        }
```
AndrewMeadows, line 182: Above is the trick for handling left-handed coordinate systems.

src/classes/HiFiAudioAPIData.ts
```diff
@@ -32,6 +32,188 @@ export class Point3D {
...
+            new Point3D({x:1, y:0, z:0}));
+        canonicalToHifi.transpose();
+
+        // The final Matrix is: canonicalToHifi * worldToCanonical 
+        this._worldToHifi = Matrix3.multiply(canonicalToHifi, worldToCanonical);
+
+        // For convenience we cache the transposed matrix for transforming in the other direction
+        this._hifiToWorld = new Matrix3(this._worldToHifi.a, this._worldToHifi.b, this._worldToHifi.c);
+        this._hifiToWorld.transpose();
+    }
```
AndrewMeadows, line 200: This is untested in TS, however I first implemented the algorithm in matlab for easy calculations and inspection and verified it worked for going from Unity left-handed coordinate frame to HiFi's right-handed system.

src/classes/HiFiAudioAPIData.ts
```diff
@@ -32,6 +32,188 @@ export class Point3D {
...
+            y: f.b.x * g.a.y + f.b.y * g.b.y + f.b.z * g.c.y,
+            z: f.b.x * g.a.z + f.b.y * g.b.z + f.b.z * g.c.z});
+        let c = new Point3D({
+            x: f.c.x * g.a.x + f.c.y * g.b.x + f.c.z * g.c.x,
+            y: f.c.x * g.a.y + f.c.y * g.b.y + f.c.z * g.c.y,
+            z: f.c.x * g.a.z + f.c.y * g.b.z + f.c.z * g.c.z});
+
+        return new Matrix3(a, b, c);
+    }
+}
```
AndrewMeadows, line 133: The **Matrix3** class was needed to do the transform calculations in **HiFiCoordinateFrameUtil**.

howard-stearns (commented):
Recording my takeaways and opinions from discussion at standup:
- I think it's generally the right thing. Nice simplification.
- Before merging, we should include a mechanism for accepting single-axis Euler rotations in the 2d case, and receiving such in the subscription. 
- On what's here now, there isn't much of a compatibility issue with existing code, as the existing cases weren't working anyway. (There may have been some specific uses that were working, though, such as a right-handed 2d system, and this would be a breaking change.)
- It would be good to include in code-comments something like the PR comment about rotations not needing transformation and why.

src/classes/HiFiAxisConfiguration.ts
```diff
@@ -4,270 +4,63 @@
...
-    static translateOrientationQuat3DToMixerSpace(axisConfiguration: HiFiAxisConfiguration, inputOrientationQuat3D: OrientationQuat3D): OrientationQuat3D {
-        let retval = new OrientationQuat3D();
-        retval = inputOrientationQuat3D;
-        return retval;
+    handedness: HiFiHandedness;
+
+    constructor(forward: Point3D, up: Point3D, handedness: HiFiHandedness) {
+        this.forward = forward;
+        this.up = up;
+        this.handedness = handedness;
```
howard-stearns, line 45: I find it slightly confusing/disturbing to have an enum in some places, and a left handed boolean in other places. (I have no preference between them.)

src/classes/HiFiAudioAPIData.ts
```diff
@@ -32,6 +32,186 @@ export class Point3D {
...
+ * to convert World-frame Position into the HiFi-frame.
+ * HiFiCoordinateFrameUtil is a helper class for this scenario.
+ *
+ * To build a HiFiCoordinateFrameUtil instance you must present three arguments:
+ * (1) The normalized FORWARD axis in the World-frame
+ * (2) The normalized UP axis in the World-frame
+ * (3) Whether it is a RIGHT- or LEFT-handed coordinate system
+ *
+ * Note: there is no need to transform Orientation into the HiFi-frame:
+ * its value remains unchanged between the two systems.
```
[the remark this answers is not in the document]
AndrewMeadows: The Peer has Position + Orientation in the World-frame.  When transforming from World-frame to HiFi-frame the Position changes but the Orientation does not.  This is true no matter how the Orientation is encoded (Quaternion, Matrix3, or even Euler Angles).

src/classes/HiFiAudioAPIData.ts
```diff
@@ -32,6 +32,186 @@ export class Point3D {
...
+            new Point3D({x:0, y:1, z:0}),
+            new Point3D({x:1, y:0, z:0}));
+        canonicalToHifi.transpose();
+
+        // The final Matrix is: canonicalToHifi * worldToCanonical 
+        this._worldToHifi = Matrix3.multiply(canonicalToHifi, worldToCanonical);
+
+        // For convenience we cache the transposed matrix for transforming in the other direction
+        this._hifiToWorld = new Matrix3(this._worldToHifi.a, this._worldToHifi.b, this._worldToHifi.c);
+        this._hifiToWorld.transpose();
```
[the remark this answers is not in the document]
AndrewMeadows: D'oh!  Thanks for pointing that out.  I gotta remember to wear my TypeScript thinking cap.

src/utilities/HiFiMath.ts
```diff
@@ -0,0 +1,593 @@
...
+// helper function that keeps an angle expressed in degrees in the range ]-360, 360[
+function sanitizeAngleDegrees(v: number): number {
+    // in the case v is Infinity or Nan,  let's special case
+    if (isNaN(v) || v === Infinity) {
+        return 0.0;
+    } else if (v === -Infinity) {
+        return -0.0;
+    } else {
+        // bring the value in the range ]-360, 360[
+        // if v is < 0 then it will cycle in ]-360, 0]
```
sabrina-shanman, line 310: Interestingly, -360 % 360 yields -0 according to my browser's javascript console (as does -720 % 360)

Although the difference is likely not meaningful, since -0 === 0 yields true.

src/utilities/HiFiMath.ts
```diff
@@ -0,0 +1,593 @@
...
+ *  The resulting rotation is defining the 'rotated' space relative to the 'base' space.
+ *  A vector Vr in "rotated' space and its equivalent value Vb in the'base' space is computed as follow:
+ *  Vb = [P][Y][R] Vr
+ * 
+ * @param euler - The euler angles.
+ * @param order - The euler order convention.
+ * 
+ * @returns The end resulting quaternion defined from the euler angles combination
+ */
+export function eulerToQuaternion(euler: OrientationEuler3D, order: OrientationEuler3DOrder): Quaternion {
```
sabrina-shanman, line 394: This, and eulerFromQuaternion below, only apply for a certain axis configuration. Maybe that should be documented.

src/classes/HiFiMixerSession.ts
```diff
@@ -442,9 +459,8 @@ export class HiFiMixerSession {
                 // We received a new orientation and updated the cache entry.
                 // Need to add the new orientation value in the newUserData
                 if (serverSentNewOrientation) {
-                    // Create the new orientation value for the newUserData and
-                    // convert the received orientation (if any) to the user space
-                    newUserData.orientationQuat = HiFiAxisUtilities.translateOrientationQuat3DFromMixerSpace(ourHiFiAxisConfiguration, userDataCache.orientationQuat);
+                    // Note: orientation remains unchanged between HiFi- and World-frame
+                    newUserData.orientationQuat = userDataCache.orientationQuat;
```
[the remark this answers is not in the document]
AndrewMeadows: The above comment is WRONG.  I performed a bad poof to myself, and then later I thought more about it and realized my mistake.  I did fix this... not sure if the fix is only in PR-124 or this one.

AndrewMeadows:
We decided to go with PR-124.  Closing this one.

PR #118, opened 2021-06-01
Signaling error handling

hd23408:
Changes from libravi https://github.com/highfidelity/libravi/pull/450

### DIFF
--- a/src/libravi/RaviCommandController.ts
+++ b/src/libravi/RaviCommandController.ts
@@ -593,8 +593,6 @@ export class RaviCommandController {
       return;
     }
 
-    RaviUtils.log("_continueProcessingListeningCommand: " + commandMessage, "RaviCommandController");
-
     // Let's try to find the matching listener(s) for the commandMessage received:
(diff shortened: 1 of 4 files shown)


howard-stearns (commented):
This does appear to be an authentic copy of the changes that have already been merged to libravi in https://github.com/highfidelity/libravi/pull/450/files

However, the base files are out of sync. For example, the new line 1227 in libravi/client/ravijs/RaviSession.ts replaces old line 1130, but in hifi-spatial-audio-js/src/libravi/RaviSession.ts it replaces old line 1093.

The discrepancy is disturbing, but I don't know what would account for the difference, or whether that difference matters. Any ideas?

hd23408:
For reference, the discrepancy was due to the fact that this PR actually pulls in multiple RAVI PRs: 
- https://github.com/highfidelity/libravi/pull/450
- https://github.com/highfidelity/libravi/pull/449
- https://github.com/highfidelity/libravi/pull/448

PR #121, opened 2021-06-03
READY FOR REVIEW: Auto-reconnect handling

hd23408:
Auto-reconnect handling, ready for testing!

The main things that are going on here:

- Additional notes available for High Fidelity employees at https://highfidelity.atlassian.net/wiki/spaces/HIFI/pages/1947369477/TypeScript+JavaScript+SDK+Reconnection+Logic

There are three new connection states (see code comments for details):
- Connecting
- Disconnecting
- Reconnecting
Additionally, `Disconnected` is now the initial state of the `HiFiCommunicator` before any connection attempts happen.

- Returned objects from HiFiCommunicator.connectToHiFiAudioAPIServer (and also informational objects that get passed into the state change handler if the handler wants 'em) now always match the new HiFiConnectionAttemptResult interface, regardless of whether the promise was resolved or rejected. These objects always have a success boolean, and then have either an error string (if rejected) or an audionetInitResponse object (if resolved). (Previously, rejected values would (often) just be a string, whereas resolved values were a JSON object with success true and audionetInitResponse filled in.) This new behavior is more consistent with how other returned objects in various other places in the code work.
 
- The `_manageConnection` method on the `HiFiCommunicator` is where the bulk of the logic around connecting and reconnecting lives.

- The `connectToHiFiAudioAPIServer` method on `HiFiCommunicator` has had the actual connecting-to-the-mixer logic pulled out into a new `_connectToHiFiMixer` method to differentiate the user's attempt to connect from internal retries

- The `_currentHiFiConnectionState` and the user's state change handler have moved onto the `HiFiCommunicator`, because it's where all of the connection/reconnection logic is happening. This is a pretty minor change in general, but it did affect the Diagnostics! So I'm passing the HiFiMixerSession a function on the HiFiCommunicator (passed as `getUserFacingConnectionState`) that will allow it to get at the user-facing state to be sure that we're logging what the user is seeing.

- In the `HiFiMixerSession`, the connection method has been rewritten to use a Promise chain rather than `await`ing the RAVI connection promises. This allows us to more closely control the behavior when we're trying to reconnect by bypassing the normal state change handlers until the connection has completely connected.

- Introduced a new `ConnectionStateChangeHandler` interface to define the messages that we're passing around between `HiFiCommunicator` and `HiFiMixerSession`. (I didn't enforce this for the user state change handler because backwards compatibility.)

(Note: This also pulls in RAVI PR 461 and 462.)


src/classes/HiFiMixerSession.ts
```diff
@@ -831,26 +836,15 @@ export class HiFiMixerSession {
...
         }
     }
 
-    /**
-     * Return the current state of the connection.
-     */
-    getCurrentHiFiConnectionState(): HiFiConnectionStates {
-        return this._currentHiFiConnectionState;
-    }
-
```
howard-stearns, line 853: I appreciate the effort to "clear the decks" and remove all the complications that we don't need.

But one reason for tracking the state (and making it available to applications), is so that an application can query the state at odd times. For example, consider the navigator property https://developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/onLine -- Some apps want to know when this changes, and so there is an event that people can listen on. But sometimes, in a handler for something else entirely, one might want to know that we are offline, and it is convenient to just ask. Yes, an application could register for changes and keep track of the last change itself, but the more untested user code that needs to be written -- by each individual application writer -- the more chance that there will be bugs. The mark of a great API isn't that it has the minimum number of API bugs, but that all the applications that use it have the minimum number of bugs.

hd23408: This method didn't get removed, it got moved up to HiFiCommunicator rather than living on HiFiMixerSession. (Users can't actually access HiFiMixerSession directly.)

howard-stearns: Excellent. (Hadn't read that code yet.)

src/classes/HiFiCommunicator.ts
```diff
@@ -392,11 +556,7 @@ export class HiFiCommunicator {
      * is still in the process of initializing its underlying HiFiMixerSession).
      */
     getConnectionState(): HiFiConnectionStates {
-        if (this._mixerSession) {
-            return this._mixerSession.getCurrentHiFiConnectionState();
-        } else {
-            return null;
-        }
+        return this._currentHiFiConnectionState;
```
hd23408, line 559: @howard-stearns This is where getConnectionState() lives nowadays. I just moved the tracking of the current connection state up a level to the communicator, so that the communicator could more easily make decisions based on it.

src/classes/HiFiCommunicator.ts
```diff
@@ -21,7 +21,21 @@ import { AvailableUserDataSubscriptionComponents, UserDataSubscription } from ".
...
+     */
+    Connecting = "Connecting",
     Connected = "Connected",
+    /**
+     * The `HiFiConnectionState` will be `"Reconnecting"` if the system is in the process of trying to
+     * automatically re-establish a pre-existing connection.
+     */
+    Reconnecting = "Reconnecting",
     Disconnected = "Disconnected",
     Failed = "Failed",
```
howard-stearns, line 40: Document the intent of these last two?

src/classes/HiFiCommunicator.ts
```diff
@@ -53,6 +67,58 @@ export enum HiFiUserDataStreamingScopes {
...
+   * connection process, regardless of the setting of this value.
+   *
+   */
+  autoRetryInitialConnection?: boolean;
+  /**
+   * The total amount of time (in seconds) to keep retrying the initial
+   * connection before giving up completely. If `autoRetryInitialConnection`
+   * is set to `true`, this defaults to 60 seconds.
+   */
+  maxSecondsToSpendRetryingInitialConnection?: number;
```
howard-stearns, line 90: Style choice:
- A single maxSecondsToSpendRetryingInitialConnection, in which a number <= 0 will not retry, and no autoRetryInitialConnection boolean.
- A autoRetryInitialConnection boolean AND a maxSecondsToSpendRetryingInitialConnection number, as shown here, in which we document (and test at runtime) that the value supplied to maxSecondsToSpendRetryingInitialConnection is a number >= 0 and give a nice error message if it isn't.

My preference is the first, but I'm not willing to die on that hill. My reasoning is:
- It isn't any simpler or clearer for the user to have the two values, because with EITHER approach, the user still needs to read through and understand about negative numbers.
- While having two parameters allows a 60 second "default in case the user picks maxSecondsToSpendRetryingInitialConnection: true", there isn't anything magically righteous about 60 seconds. I'd rather have the application developer think about what kind of timeout is appropriate for their application.
- It takes more "space" to capture the state of the configuration, e.g., in logging the configuration to console or to HiFi for the purpose of examining connection failures.

src/classes/HiFiCommunicator.ts
```diff
@@ -233,10 +324,6 @@ export class HiFiCommunicator {
      * Please reference our {@link https://www.highfidelity.com/api/guides/misc/getAJWT|"Get a JWT" guide} for additional context.
      * 4. Pass the created JWT to `connectToHiFiAudioAPIServer()`.
      * 
-     * As of 2021-01-21, we've added code in this function which, in the browser context, searches for a `token` URL query parameter and, if a JWT
-     * isn't supplied as an argument to this function, uses the value of that `token` URL query parameter as the JWT.
-     * We should remove that later, because we almost certainly don't want this to stay in the API code, but it's _very_ convenient for sample apps for right now.
-     *
```
howard-stearns, line 239: Thank you!

src/classes/HiFiCommunicator.ts
```diff
@@ -305,6 +406,78 @@ export class HiFiCommunicator {
...
+            // should probably not trigger a change (they don't in the current implementation), because "UNAVAIALBLE" is
+            // itself a failure status (for "velvet rope"). However, if the change from "UNAVAILABLE" goes to something else (e.g. "CONNECTING")
+            // then the change _should_ be triggered. Need to think through that a little bit more, and also re-evaluate how that behavior has
+            // changed since the websocket signaling changes were implemented (the UNAVAILABLE state gets communicated the same way, which
+            // is what made its original handling logic so very complicated before RaviSession was updated with the new short-circuit approach).
+            // It's probably easier now (i.e. I think the RaviSession may just stay at UNAVAILABLE, in which case this would all be moot).
+            this._currentHiFiConnectionState = state;
+            if (this.onConnectionStateChanged) {
+                this.onConnectionStateChanged(this._currentHiFiConnectionState);
+            }
```
howard-stearns, line 477: How will the user's onConnectionStateChanged handler interact with our stuff (the TODO). E.g., which gets called first? Can one change the state as seen by the other?

src/classes/HiFiCommunicator.ts
```diff
@@ -305,6 +406,78 @@ export class HiFiCommunicator {
...
+                  AND try to reconnect
+                  ELSE handle the new state as implemented below
+
+               2. ELSE IF the new state is CONNECTED
+                  THEN handle the new state as implemented below
+                  AND kill the maxSecondsToSpendRetryingOnDisconnect timeout if it's set
+
+               4. ELSE IF the new state is other (CONNECTING or RECONNECTING)
+                  AND IF the new state is not the same as this._currentHiFiConnectionState
+                  THEN handle the new state as implemented below
```
howard-stearns, line 463: There are complex bits of external state that interact with all of this. My instinct is to NOT try to work through these now, but I'll raise it to give folks a chance to argue otherwise:
- How does this interact with being offline? Is that the same as our server reporting "unavailable" for capacity? Should we set up handler to "try again" when one comes back online, or is that for the app to decide (because it may involve asking the user if that's what they want), or does the answer depend on how long we were off line, or on what our _currentHiFiConnectionState was when we went offline?
- Same question for the page not being visible, and is that different for desktop v mobile.

src/classes/HiFiCommunicator.ts
hd23408: My intent was to follow our logic first (including kicking off retries as needed), and then call the user's onConnectionStateChanged handler _only if_ there is an actual change in state after following our logic. 

That would happen regardless of whether the change in state was the result of a change bubbled up from the webrtc session or was due to a 'conscious' decision on our part. For instance, if the underlying webrtc session had been connected, and then it disconnected (which triggered _our_ handler), and then we immediately started retrying, the TODO logic would explicitly change the overall Communicator's state to "RECONNECTING" and trigger the user's handler with "RECONNECTING" as the user-facing state, so that the user could put up a "we're trying to reconnect you" message if they want to. If that reconnect attempt failed and we tried to reconnect again, we wouldn't call the user's handler a second time with the same "RECONNECTING" state. However, if we ultimately timed out the attempts to reconnect, we would then switch to "FAILED" and call the user's handler.

The user's connection handler (or any user code) could hypothetically change HiFiCommunicator._currentHiFiConnectionState directly if they really wanted to, because TypeScript private doesn't prevent JavaScript from messing with things. (Unless we change stuff and use new JS features: https://www.valentinog.com/blog/private/ ) But that's JavaScript for you.

src/classes/HiFiCommunicator.ts
hd23408: Yeah, I agree with your instinct on this! To answer the direct question, if the user goes offline it will manifest as a dropped connection first, and then an "unable to make signaling connection" error for any retries, because the WebSocket attempt will immediately fail.

But I think that we should not set up handlers to watch connection state and make decisions based on that -- instead, simply retry as instructed for the amount of time instructed, and if it still can't connect just move to a failed state and stop trying.

src/classes/HiFiCommunicator.ts
hd23408: 👍

src/classes/HiFiCommunicator.ts
```diff
@@ -53,6 +96,70 @@ export enum HiFiUserDataStreamingScopes {
...
+
+  /**
+   * Whether or not to automatically attempt to reconnect if an existing
+   * connection is disconnected. When this is set to true, we will attempt
+   * to reconnect if any disconnect from any cause occurs.
+   * By default, reconnections are not automatically attempted.
+   * NOTE: The retrying that happens when this is set to `true` does not currently take into account
+   * the reason WHY a connection was disconnected. This means that if this is
+   * set to true, a connection that is disconnected via a purposeful server-side
+   * action (e.g. a "kick") will be automatically reconnected. (However, connections
```
howard-stearns, line 130: Are we creating a bug? What overall system behavior are we/users expecting for kick?

src/classes/HiFiMixerSession.ts
```diff
@@ -852,86 +887,94 @@ export class HiFiMixerSession {
...
-    _setCurrentHiFiConnectionState(state: HiFiConnectionStates): void {
-        if (this._currentHiFiConnectionState !== state) {
-            this._currentHiFiConnectionState = state;
+    async _onConnectionStateChange(state: HiFiConnectionStates, result: ConnectionAttemptResult): Promise<void> {
+        if (this.getCurrentHiFiConnectionState() !== state) {
             if (this.onConnectionStateChanged) {
-                this.onConnectionStateChanged(this._currentHiFiConnectionState);
+                this.onConnectionStateChanged(state, result);
             }
             this._hifiDiagnostics.fire();
```
howard-stearns, line 920: The intention with diagnostics is that we prime (both hifi and ravi diagnostics) on connect, and the fire hifiDiagnostics on anything else here.  That happened implicitly (without so much as a code comment, alas) before, with the state changes as they were.

Now that there are more states, one could guard this `fire()` here, but it would probably make the code clearer to _remove two `prime` lines (new lines 308, 309) in the audionet.init handler_,  and change this `fire()` line here to:
```
            if (state === HiFiConnectionStates.Connected) {
                this._raviDiagnostics.prime(this.mixerInfo.visit_id_hash);
                this._hifiDiagnostics.prime(this.mixerInfo.visit_id_hash);
            } else {
                this._hifiDiagnostics.fire();
            }
```

src/libravi/RaviSignalingConnection.ts
```diff
@@ -434,8 +455,22 @@ class RaviSignalingWebSocketImplementation {
         signalingConnection._handleStateChange({}, RaviSignalingStates.CLOSED);
```
howard-stearns, line 455: Is there any chance to get here? If so, don't we want all the same things to happen, in terms of setting _webSocket to null and removing event listeners?

src/libravi/RaviSignalingConnection.ts
```diff
@@ -396,24 +396,45 @@ class RaviSignalingWebSocketImplementation {
...
+  _onWebSocketClose = (function(event: any) {
+    if (event && event.code && event.code > 4000) {
+      // This "close" event is really an error, because we're
+      // returning one of our custom error codes. Treat it as such.
+      RaviUtils.err("_handleStateChange: signaling error code " + event.code + ":  " + event.reason, "RaviSignalingConnection");
+      this._raviSignalingConnection._handleStateChange(event, RaviSignalingStates.ERROR);
+    } else {
+      this._raviSignalingConnection._handleStateChange(event, RaviSignalingStates.CLOSED);
+    }
+  }).bind(this);
```
howard-stearns, line 436: A matter of style: This code is completely legit, but there are a few (debatable) reasons why we might want to just have the constructor set this.onWebSocketClose = function(...etc) {...};, without the bind.:
- Most Javascript programmers don't know exactly what bind does, nor the difference between function and fat arrow.  Elsewhere we have made the specific decision to not used advanced features that programmers might not know, even when it would make the code clearer. (In this case, the code is in a private method, so arguably that's ok.)
- Risk of platform problems. As far as I know, all browsers that we target support _instance_ variables initialized with this syntax (regardless of whether they are functions), but I happen to know that Safari doesn't support static/_class_ variables to be initialized this way. So I'm a little nervous about this one.

howard-stearns: Update: CURRENT safari supports this syntax for both instance AND class vars, which makes me wonder (and unable to test) whether older (but recent) safari had indeed supported this syntax for instance variables!

src/classes/HiFiMixerSession.ts
```diff
@@ -186,16 +188,25 @@ export class HiFiMixerSession {
...
      * Contains information about the mixer to which we are currently connected.
      */
     mixerInfo: any;
+    
+    /**
+     * Track whether or not we're in the process of attempting to connect. (This is
+     * a little bit hacky, but given the maze of twisty callbacks and Promises, it
+     * seems the most sane way to avoid double-connecting.)
+     */
+    _tryingToConnect: boolean;
```
howard-stearns, line 204: I lost track: in the end, did it turn out that we always go to state Connecting when we're trying to connect? (In which case why do we need this as well?)

Or maybe that Connecting means we're generally connecting, retrying, waiting for a retry, etc., while this is really a micro state means that we're not in a waiting between attempts, but have specifically initiated an attempt?  (In which case, let's include a comment to that effect?)

src/classes/HiFiCommunicator.ts
```diff
@@ -159,13 +300,21 @@ export class HiFiCommunicator {
         if (onUsersDisconnected) {
             this.onUsersDisconnected = onUsersDisconnected;
         }
+        if (onConnectionStateChanged) {
+            this.onConnectionStateChanged = onConnectionStateChanged;
+        }
+        this._connectionRetryAndTimeoutConfig = HiFiConstants.DEFAULT_CONNECTION_RETRY_AND_TIMEOUT;
+        if (connectionRetryAndTimeoutConfig) {
+            Object.assign(this._connectionRetryAndTimeoutConfig, connectionRetryAndTimeoutConfig);
```
howard-stearns, line 308: This is a bug. It is side effecting the internal contents of the object at HiFiConstants.DEFAULT_CONNECTION_RETRY_AND_TIMEOUT, which will effect other HiFiCommunicator instances.

I think I would specify either HiFiConstants.DEFAULT_CONNECTION_RETRY_AND_TIMEOUT or the empty {} as the default value for the connectionRetryAndTimeoutConfig parameter of the constructor, and then replace all four lines here with:
```
this._connectionRetryAndTimeoutConfig = {}; // Or uses class field syntax if we're willing to require latest Safari.
Object.assign(this._connectionRetryAndTimeoutConfig, HiFiConstants.DEFAULT_CONNECTION_RETRY_AND_TIMEOUT, connectionRetryAndTimeoutConfig);
```

src/libravi/RaviSignalingConnection.ts
hd23408: Yup, agree -- I think this _should_ just be `if (! this._websocket) return etc.` (to catch the "this._websocket is already null" case) and let the `if (this._webSocket.readyState === crossPlatformWebSocket.CLOSED)` fall through to the rest of the stuff that this method does.

src/libravi/RaviSignalingConnection.ts
hd23408: I don't feel particularly strongly one way or the other, I just personally like `bind`s. :-) Happy to change if you think it's necessary!

src/classes/HiFiMixerSession.ts
hd23408: The latter -- Connecting and Reconnecting mean that we're generally "in the greater process of trying to connect" and this is a micro state. I will add a comment!

src/classes/HiFiCommunicator.ts
hd23408: Excellent question! This came up back when we were initially talking about autoreconnects. I would expect that people are thinking that "kick" would not trigger a client-side autoreconnect, and we https://highfidelity.atlassian.net/browse/HIFI-630 was filed for having the server explicitly tell the client when a kick was happening so that it could trap kicks. For now I think it's probably acceptable to document that "autoreconnects and kicks are incompatible" in the spirit of making incremental changes. I don't think it _quite_ qualifies as a "bug", but it does need to be aggressively documented for sure.

src/classes/HiFiCommunicator.ts
```diff
@@ -159,13 +300,21 @@ export class HiFiCommunicator {
...
 
         this._mixerSession = new HiFiMixerSession({
             "userDataStreamingScope": userDataStreamingScope,
             "onUserDataUpdated": (data: Array<ReceivedHiFiAudioAPIData>) => { this._handleUserDataUpdates(data); },
             "onUsersDisconnected": (data: Array<ReceivedHiFiAudioAPIData>) => { this._onUsersDisconnected(data); },
-            "onConnectionStateChanged": onConnectionStateChanged,
-            "onMuteChanged": onMuteChanged
+            "onConnectionStateChanged": (state: HiFiConnectionStates, message: ConnectionAttemptResult) => { this._manageConnection(state, message); },
+            "onMuteChanged": onMuteChanged,
+            "getUserFacingConnectionState": this.getConnectionState.bind(this)
```
howard-stearns, line 317: I didn't follow what this is a about (or for that matter, what getConnectionState is for as a function rather than just letting the state property be public, as it seems to only be used here, but I guess that's a separate question).

It looks like maybe the only reason for this is to decide whether to call the  onConnectionStateChanged that we pass in here, but I don't understand why not just alway call it and let the manager decide what to do about it. After all, the manager already has access to that info.

src/classes/HiFiCommunicator.ts
hd23408: Good eye -- thank you!!

src/classes/HiFiCommunicator.ts
hd23408: Let's discuss "in person" -- I also am not prepared to die on the hill, but it was easier for me just to move forward with what I had. Would be easy enough to change at this point though!

src/classes/HiFiCommunicator.ts
hd23408: Ah -- the main reason for this is that:
1) The connection state is now tracked on the HiFiCommunicator, but
2) The diagnostics code only knows about the HiFiMixerSession, not the HiFiCommunicator, and
3) The HiFiMixerSession doesn't have a reference to the HiFiCommunicator that's using it, so
4) In order for the diagnostics code to be able to find out what the current "user facing" state is, it needs to be able to get that from the HiFiMixerSession (since that's all it knows about), so
5) The HiFiMixerSession needs to be able to get it from the HiFiCommunicator, so
6) This is a way for the HiFiMixerSession to get the connection state that the HiFiCommunicator knows about without having to pass a reference to the ENTIRE HiFiCommunicator into the HiFiMixerSession.

src/classes/HiFiMixerSession.ts
hd23408: @howard-stearns I made this change and will resolve this comment but please let me know if I didn't do it right!

src/classes/HiFiCommunicator.ts
howard-stearns: If this is here JUST for the diagnostics code....  I'd rather have the session/state/management code be clean than to have the diagnostics code be clean. (I could create/expose some hooks in the MixerSession, and poke those hooks from some temporary code in the Communicator.)

hd23408: Yep, it's there just for the diagnostics code. Both because the diagnostics code needs to get the "current user-visible state" and also because HiFiMixerSession's `_onConnectionStateChange` method needs to be able to check to see if the state being passed is different than the current state (so that it can decide when/if to fire the diagnostics).

I mean TBH I actually think this is a reasonably elegant way of passing the information down (I was proud of myself for thinking of it 😄 ) , but I'm also fine with your exposing temporary hooks to store a copy of the info in the MixerSession instead. Do you want to flag this and rework it post-merge, or add a commit to this PR? (Or I can give it a shot if you'd like.)

howard-stearns: I think we can clean it up later. The main thing is to understand what the design is.

howard-stearns: Sorry for having made this work more complicated!

hd23408: Not at all! I'll add a comment in that general area with a TODO about cleaning it up later.

hd23408: (Comment added to the HiFiMixerSession's parameter explanation. Also added comments for other HiFiMixerSession parameters that were previously uncommented, while I was in there.)

PR #124, opened 2021-06-07
More complete overhaul of coordinate-frame config and orientation helpers

AndrewMeadows:
This is a more complete and drastic version of the work in PR-114.  I'm creating it as a separate PR for comparison between the levels of changes.

In particular in this PR: all the `OrientationEuler3D` and `OrientationEuler3DOrder` stuff has been **removed**.  In its place are a reduced set of utils that can extact the **yaw** (rotation about vertical axis, aka "azumuth"), **pitch** (rotation up from horizontal plane, aka "elevation"), or the full **yawPitchRoll** decomposition from an `OrientationQuat3D`.  What has been removed (for the benefit of humanity) are the myriad variations of Euler angle decompositions for which: **(a)** no one is expected to use and **(b)** no one should be encouraged to use.


src/classes/HiFiCommunicator.ts
```diff
@@ -90,6 +91,18 @@ export class HiFiCommunicator {
...
+     * If the World coordinate system is NOT compatible with the HiFi coordindate frame used by the mixer              
+     * then configure a HiFiCoordinateFrameUtil to transform to and from HiFi-frame.
+     *
+     * The World-frame is compatible iff:  
+     * (1) It is right-handed
+     * (2) It uses the Y-axis (positive or negative, doesn't matter) for the UP direction.
+     *           
+     * For all other cases create a {@link WorldFrameConfiguration} and pass it to the HiFiCommunicator constructor.
+     */
+    private _coordFrameUtil?: HiFiCoordinateFrameUtil;
```
AndrewMeadows, line 104: It would be up to the Application dev to supply a `_coordFrameUtil` to the `HiFiCommunictor` or not.  If one is supplied then it will be passed to each `HiFiMixerSession` constructor.

src/classes/HiFiMixerSession.ts
```diff
@@ -199,14 +212,15 @@ export class HiFiMixerSession {
...
-    constructor({ userDataStreamingScope = HiFiUserDataStreamingScopes.All, onUserDataUpdated, onUsersDisconnected, onConnectionStateChanged, onMuteChanged }: { userDataStreamingScope?: HiFiUserDataStreamingScopes, onUserDataUpdated?: Function, onUsersDisconnected?: Function, onConnectionStateChanged?: Function, onMuteChanged?: OnMuteChangedCallback }) {
+    constructor({ userDataStreamingScope = HiFiUserDataStreamingScopes.All, onUserDataUpdated, onUsersDisconnected, onConnectionStateChanged, onMuteChanged, coordFrameUtil }: { userDataStreamingScope?: HiFiUserDataStreamingScopes, onUserDataUpdated?: Function, onUsersDisconnected?: Function, onConnectionStateChanged?: Function, onMuteChanged?: OnMuteChangedCallback, coordFrameUtil?: HiFiCoordinateFrameUtil}) {
         this.webRTCAddress = undefined;
         this.userDataStreamingScope = userDataStreamingScope;
         this.onUserDataUpdated = onUserDataUpdated;
         this.onUsersDisconnected = onUsersDisconnected;
         this._mixerPeerKeyToStateCacheDict = {};
         this._lastSuccessfulInputAudioMutedValue = false;
         this.onMuteChanged = onMuteChanged;
+        this._coordFrameUtil = coordFrameUtil;
```
AndrewMeadows, line 223: The `coordFrameUtil` is passed in at the `HiFiMixerSession` ctor.  If it exists then it will be used to transform to/from the HiFi-coordinate-frame as necessary.

src/utilities/HiFiMath.ts
```diff
@@ -0,0 +1,460 @@
...
+        let x = b.x - a.x;
+        let y = b.y - a.y;
+        let z = b.z - a.z;
+        return x*x + y*y + z*z;
+    }
+}
+
+/**
+ * A Quaternion for 3D rotations.
+ */
```
AndrewMeadows, line 77: A nearly complete `Quaternion` implementation was required, mostly so that we could supply unit-tests to verify the more customer-centric utils are working correctly.

src/utilities/HiFiMath.ts
```diff
@@ -0,0 +1,460 @@
...
+        let pitchedUp = Vector3.cross(yawedRight, rotatedForward); // note, this is already unitary
+        let roll = Math.atan2( Vector3.dot(rotatedUp, yawedRight), Vector3.dot(rotatedUp, pitchedUp));
+
+        // convert all angles to degrees
+        yaw *= RADIANS_TO_DEGREES;
+        pitch *= RADIANS_TO_DEGREES;
+        roll *= RADIANS_TO_DEGREES;
+
+        return {"yaw": yaw, "pitch": pitch, "roll": roll};
+    }
```
AndrewMeadows, line 194: For application developers who really want Euler angles: the utility `Quaternion.getEulerAngles()` method is available.  This provides the most likely variety of Euler angle decomposition they would want, which is a yaw, pitch, and roll (in that order) about the local up, right, and forward axes **(+)**.  Other more obscure varieties of Euler angles only add confusion, and we shouldn't expect sane customers to want them.  Those that do should be required to write their own translation tools for such follies.

**(+)** One cannot simply say "Euler angles" and be 100% clear.  One must specify details of the Euler angle decomposition for mathematical rigor, else risk bugs and frustration.

src/utilities/HiFiMath.ts
```diff
@@ -0,0 +1,460 @@
...
+            let rotatedRight = this.rotateVector(right);
+            projectedLengthSquared = rotatedRight.x * rotatedRight.x + rotatedRight.z * rotatedRight.z;
+            azimuth = Math.acos(Vector3.dot(rotatedRight, right) / Math.sqrt(projectedLengthSquared)); 
+            if (Vector3.dot(rotatedRight, forward) < 0.0) {
+                azimuth *= -1;
+            }
+        }
+        // finally, convert from radians to degrees
+        return azimuth * RADIANS_TO_DEGREES;
+    }
```
AndrewMeadows, line 225: One special case is: 2D applications.  These typically only expect and care about rotations about the vertical axis.  This is the "azimuth", aka the "yaw".  For this common case we offer two equivalent helper methods:  `Quaternion.getAzimuth()` and `Quaternion.getYaw()`.  2D customers would be required to use those on the incoming OrientationQuat3D supplied inside the HiFiAudioAPIData.orientation field.  And for outgoing data the customer must use: `orientation = Quaternion.fromEulerAngles({yaw: angle});`.

package.json
```diff
@@ -55,6 +55,7 @@
   },
   "dependencies": {
     "esm": "^3.2.25",
+    "npm": "^7.16.0",
```
howard-stearns, line 58: If we go with this, let's remember to remove this change. See https://highfidelity.slack.com/archives/C0XU8ESJC/p1627948590001600

zfox23 (changes requested):
Excellent work. Thank you, Andrew.

I'm going to start writing down comments as I'm using these changes as I'm modifying the documentation and sample apps.

**EDIT: Please ignore the below; I now believe that the bug I thought I found was with client code and not with library code or with this PR's code. I'm keeping the below for historical purposes.**

First thing I've noticed: When I call `HiFiCommunicator.updateUserDataAndTransmit()`, and one or more of the Quaternion components hasn't changed, the orientation of this client on the server doesn't change. Example return value of `updateUserDataAndTransmit()`:

```
{success: true, rawDataTransmitted: "{\"W\":923.8795325112865,\"X\":382.6834323650897,\"Z\":0}"}
```

Notice how there's no "Y" component in there. When I modify the library code to transmit WXYZ anytime any one Quat component changes, then I get the results I expect:

```
{success: true, rawDataTransmitted: "{\"W\":923.8795325112867,\"X\":-382.6834323650898,\"Y\":0,\"Z\":0}"}
```

Here is the diff associated with the library code change:
```
diff --git a/src/classes/HiFiMixerSession.ts b/src/classes/HiFiMixerSession.ts
index 5a73b37..0573acf 100644
--- a/src/classes/HiFiMixerSession.ts
+++ b/src/classes/HiFiMixerSession.ts
@@ -1178,19 +1178,10 @@ export class HiFiMixerSession {
                 }

                 // The mixer expects Quaternion to be mulitiplied by 1000.
-                if (changedComponents.w) {
-                    dataForMixer["W"] = translatedOrientation.w * 1000;
-                }
-                if (changedComponents.x) {
-                    dataForMixer["X"] = translatedOrientation.x * 1000;
-                }
-                if (changedComponents.y) {
-                    dataForMixer["Y"] = translatedOrientation.y * 1000;
-                }
-                // if (changedComponents.z) {
-                // Need to send Z all the time at the moment until we merge the fix https://github.com/highfidelity/audionet-hifi/pull/271
+                dataForMixer["W"] = translatedOrientation.w * 1000;
+                dataForMixer["X"] = translatedOrientation.x * 1000;
+                dataForMixer["Y"] = translatedOrientation.y * 1000;
                 dataForMixer["Z"] = translatedOrientation.z * 1000;
-                //}
             }
         }
```

This indicates to me that it's necessary to send all Quat components all the time, even if one or more of them does not change. Alternatively, we need to make sure the mixer accepts differential changes, which I believe was the intention. I think this has probably been a bug in the system for several months now.

I notice a comment in the library code:
```
// Need to send Z all the time at the moment until we merge the fix https://github.com/highfidelity/audionet-hifi/pull/271
```

Perhaps relevant?

howard-stearns:
@roxanneskelly - Refresh my memory (and see comment above). I think that the mixer expects all the quaternion properties (x,y,z,w) to be set. 

The guard code that I fixed in the above reference PR was meant to make the server not crash if a property is omitted, but it is n to meant to provide a default value. 

I'm not sure that it is meaningful to default unsupplied values. What would the default be? Zero? I.e., not a server bug. If there is a meaningful default we could make an enhancement -- e.g., maybe if any of x, y, z, or w is supplied as non-zero, the others are defaulted to 0, and then the whole thing normalized? (Yuck?) But without such, I think the client needs to send all properties if it sends any. If it doesn't, then that would be a client bug. (I think.)

src/utilities/HiFiMath.ts
```diff
@@ -0,0 +1,464 @@
...
+        let yawedRight = new Vector3({ x: Math.cos(yaw), y: 0.0, z: -Math.sin(yaw) });
+        let pitchedUp = Vector3.cross(yawedRight, rotatedForward); // note, this is already unitary
+        let roll = Math.atan2( Vector3.dot(rotatedUp, yawedRight), Vector3.dot(rotatedUp, pitchedUp));
+
+        // convert all angles to degrees
+        yaw *= RADIANS_TO_DEGREES;
+        pitch *= RADIANS_TO_DEGREES;
+        roll *= RADIANS_TO_DEGREES;
+
+        return {"yaw": yaw, "pitch": pitch, "roll": roll};
```
zfox23, line 193: Here (and potentially elsewhere similar), please change the key names for this object to include units, i.e.:

```js
        return {"yawDegrees": yaw, "pitchDegrees": pitch, "rollDegrees": roll};
```

AndrewMeadows:
It should be possible for the client to only send the changed components of a Quaternion, however in general a meaningful Quaternion cannot be assembled from partial components: to stuff them into a Quaternion with any default values would effectively produce corrupt data.  The changed components must be stored as such until they are applied to the previous Quaternion cached on the receiving end.

Note: in the special 2D case where the User only ever rotates about the UP axis then only two of the Quaternion components will ever be non-zero: W and whatever component corresponds to the UP axis.  In our case I believe the hifi spatial audio uses the +Y axis as UP.

I did not see that comment about always sending the Z-axis.  If the comment is to be believed then we can stop doing it because PR-271 has been merged already.

PR #137, opened 2021-06-22
Switch pako from require to import

nc:
For easier use with Rollup and Vitejs which by default handle imported code far better.


hd23408 (approved):
Awesome! Thank you, Namit!

PR #155, opened 2021-07-02
MixerSession connect is no longer async, and update MediaStream to work in node context

hd23408:
To fix tests!


src/classes/HiFiMixerSession.ts
```diff
@@ -639,12 +653,15 @@ export class HiFiMixerSession {
...
             this._onConnectionStateChange(HiFiConnectionStates.Failed, { success: false, error: error });
+            return false;
         })
         .finally(() => {
             this._raviSignalingConnection.removeStateChangeHandler(tempUnavailableStateHandler);
             this.concurrency = 0;
             this._tryingToConnect = false;
         });
+
+        return true;
```
howard-stearns, line 664: I guess the promise itself is evaluated for effect, and the value not used anywhere? 
So it's intended to return true, and not the promise?

howard-stearns: And the doc string says it returns a boolean indicating success or failure, but we don't actually know the success at this point in time, right?

hd23408: This is why I needed more eyes on it! :-) Thank you, @howard-stearns .  I've moved the `return true` inside the last `then`, and at that point we SHOULD know that it was successful. So I believe that with this _latest_ change, it is appropriate for `HiFiMixerSession.connectToHiFiMixer()` to be considered synchronous, and the boolean it returns should, in fact, indicate success or failure. (I think.) Does that sound right?

hd23408: @howard-stearns Thank you again so much for your help with this! `connectToHiFiMixer()` now returns `void` with an explicit note about getting the "return value" ONLY through the use of callbacks. The test has been updated to use a callback and check to make sure it gets called with the appropriate values. Whew.

tests/integration/serverConnections.integration.test.ts
```diff
@@ -127,6 +127,7 @@ describe('Mixer connections', () => {
     describe('Connections that should be denied', () => {
         beforeEach(() => {
             hifiCommunicator = new HiFiCommunicator();
+            jest.setTimeout(30000);
```
hd23408, line 130: These timeouts allow time for auto-reconnects to finish up when we're defaulting auto-reconnects to "true"

howard-stearns (approved):

PR #166, opened 2021-08-06
Patch version update to v1.4.1

bridie-hifi:
Patch version update to v1.4.1

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hifi-spatial-audio",
-  "version": "1.4.1-5",
+  "version": "1.4.1",
   "description": "The High Fidelity Audio Client Library allows developers to integrate High Fidelity's spatial audio technology into their projects.",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hifi-spatial-audio",
-  "version": "1.4.1-5",
+  "version": "1.4.1",
```
howard-stearns, line 3: I don't know what we meant by 1.4.1-5 .... that's not a semver format that I recognize.
But in any case, wouldn't the next one be 1.4.2?

bridie-hifi: IDK why the patch version changes with each (?) commit, but this change correctly moves us from the CURRENT released version (1.4.0 - https://github.com/highfidelity/hifi-spatial-audio-js/releases/tag/v1.4.0) to 1.4.1.

howard-stearns (approved):

PR #168, opened 2021-08-11
Clear peers on disconnect

AndrewMeadows:
This PR fixes bug HIFI-847 in the JS API.
https://highfidelity.atlassian.net/browse/HIFI-847

This work is a theoretical fix: I haven't been able to test it yet.

### DIFF
--- a/src/classes/HiFiMixerSession.ts
+++ b/src/classes/HiFiMixerSession.ts
@@ -152,7 +152,7 @@ export class HiFiMixerSession {
      * 
      * Thus, the Library user should never have to care about the `_mixerPeerKeyToStateCacheDict`.
      */
-    private _mixerPeerKeyToStateCacheDict: any;
+    private _mixerPeerKeyToStateCacheDict: { [id: string] : ReceivedHiFiAudioAPIData };
 
     /**
      * We will track whether or not the input stream is stereo, so that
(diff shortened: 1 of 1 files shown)


src/classes/HiFiMixerSession.ts
```diff
@@ -370,13 +370,13 @@ export class HiFiMixerSession {
...
-                            deletedUserData.providedUserID = this._mixerPeerKeyToStateCacheDict[mixerPeerKey].providedUserID;
+                for (const key of mixerPeerKeys) {
+                    let peerData = this._mixerPeerKeyToStateCacheDict[key];
+                    if (peerData.hashedVisitID === hashedVisitID) {
+                        if (peerData.providedUserID) {
+                            deletedUserData.providedUserID = peerData.providedUserID;
                         }
-                        // TODO: remove the entry from the peer state cache -- is this OK?
-                        //delete this._mixerPeerKeyToStateCacheDict[mixerPeerKey];
+                        delete this._mixerPeerKeyToStateCacheDict[key];
```
AndrewMeadows, line 379: While looking over the code I noticed this **for** loop was performing double map lookups in **_mixerPeerKeyToStateCacheDict** in the common case and triple lookups in the uncommon.  This offends my sensibilities for high performance code and so I was compelled to fix it.

src/classes/HiFiMixerSession.ts
```diff
@@ -699,6 +699,7 @@ export class HiFiMixerSession {
         await close(this._raviSignalingConnection, "Signaling Connection", RaviSignalingStates.CLOSED);
         await close(this._raviSession, "Session", RaviSessionStates.CLOSED);
 
+        this._clearPeerData();
```
AndrewMeadows, line 702: Here is the real fix for HIFI-847: clear the peer cache on disconnect.

src/classes/HiFiMixerSession.ts
AndrewMeadows: Oh yeah, also notice that I did uncomment the code for the old **TODO**.  Yes, it is a good idea to actually purge old peer data from the cache when it gets deleted.

src/classes/HiFiMixerSession.ts
```diff
@@ -1231,6 +1232,25 @@ export class HiFiMixerSession {
...
+                if (peerData.providedUserID) {
+                    deletedUserData.providedUserID = peerData.providedUserID;
+                }
+                allDeletedUserData.push(deletedUserData);
+            }
+            this.onUsersDisconnected(allDeletedUserData);
+        }
         this._mixerPeerKeyToStateCacheDict = {};
     }
 }
```
AndrewMeadows, line 1256: Note: we aren't just clearing the peer cache here.  If **onUserDisconnect** callback exists we also assemble **allDeletedUserData** and feed it to **onUserDisconnect**.  This tells the Application to remove any stale peer data it happens to be using.

src/classes/HiFiMixerSession.ts
```diff
@@ -358,39 +358,6 @@ export class HiFiMixerSession {
...
-                allDeletedUserData.push(deletedUserData);
-            }
-
-            // TODO: remove the entry from the peer state cache
-            this.concurrency -= allDeletedUserData.length;
-            if (this.onUsersDisconnected && allDeletedUserData.length > 0) {
-                this.onUsersDisconnected(allDeletedUserData);
-            }
-        }
-
```
AndrewMeadows, line 393: I've moved the handling of **deleted_visit_ids** down below the handling of **peers**.  It is theoretically possible for a change to exist in peers for someone who was also deleted in the same frame, and when a change is not found in the map the peer is added to it.  By handling the deletions last we can avoid re-adding the changed user after deletion.

src/classes/HiFiMixerSession.ts
```diff
@@ -536,6 +503,38 @@ export class HiFiMixerSession {
...
+                });
+
+                let mixerPeerKeys = Object.keys(this._mixerPeerKeyToStateCacheDict);
+                for (const key of mixerPeerKeys) {
+                    let peerData = this._mixerPeerKeyToStateCacheDict[key];
+                    if (peerData.hashedVisitID === hashedVisitID) {
+                        if (peerData.providedUserID) {
+                            deletedUserData.providedUserID = peerData.providedUserID;
+                        }
+                        delete this._mixerPeerKeyToStateCacheDict[key];
```
AndrewMeadows, line 525: Note: I'm correctly deleting the peer from the map on delete.  Previously this operation was commented out with a TODO comment about it (see the above removed block).

howard-stearns: thank you!

howard-stearns (approved):

PR #177, opened 2021-09-01
Allow per-user attenuation,  rolloff, and volume threshold to be NaN, in which case the value for the zone/space is used

sabrina-shanman:


src/classes/HiFiAudioAPIData.ts
```diff
@@ -44,6 +44,12 @@ export class HiFiAudioAPIData {
      * A volume level below this value is considered background noise and will be smoothly gated off.
      * The floating point value is specified in dBFS (decibels relative to full scale) with values between -96 dB (indicating no gating)
      * and 0 dB (effectively muting the input from this user). It is in the same decibel units as the VolumeDecibels component of UserDataSubscription.
+     * Setting this value to `NaN` or `null` will cause the volume threshold from the space to be used instead.
+     * 
+     * **COMPATIBILITY WARNING:** In the future, the High Fidelity Audio API server will only fall back to the space volume threshold
+     * if the threshold is `NaN`.
+     *
+     * If you don't supply a `volumeThreshold` when constructing instantiations of this class, it will be ignored.
```
howard-stearns, line 52: I didn't follow what this sentence means, here and below.

For example, does not setting it mean that the default for the space will be used?

src/classes/HiFiAudioAPIData.ts
```diff
@@ -44,6 +44,12 @@ export class HiFiAudioAPIData {
      * A volume level below this value is considered background noise and will be smoothly gated off.
      * The floating point value is specified in dBFS (decibels relative to full scale) with values between -96 dB (indicating no gating)
      * and 0 dB (effectively muting the input from this user). It is in the same decibel units as the VolumeDecibels component of UserDataSubscription.
+     * Setting this value to `NaN` or `null` will cause the volume threshold from the space to be used instead.
+     * 
+     * **COMPATIBILITY WARNING:** In the future, the High Fidelity Audio API server will only fall back to the space volume threshold
+     * if the threshold is `NaN`.
```
howard-stearns, line 50: I didn't follow what this sentence means. It seems to imply that currently (i.e., not yet the future behavior), there is some way to fall back to the space volume threshold other than setting the value to NaN, and that in the future, .... ?

sabrina-shanman: Does the sentence above the compatibility warning help clarify, or is it still confusing?

Is it just this particular compatibility warning that is confusing, or are the others confusing as well?

src/classes/HiFiAudioAPIData.ts
sabrina-shanman: I understand what you mean. I will try to clarify.

src/classes/HiFiMixerSession.ts
```diff
@@ -1208,19 +1208,19 @@ export class HiFiMixerSession {
         }
 
         if (typeof (currentHifiAudioAPIData.volumeThreshold) === "number" ||
-            currentHifiAudioAPIData.volumeThreshold === null) {
+            currentHifiAudioAPIData.volumeThreshold === null) { // May be NaN
```
howard-stearns, line 1211: The comment is confusing on this line. Maybe it should be on line 1210?

src/classes/HiFiCommunicator.ts
```diff
@@ -1012,16 +1022,16 @@ export class HiFiCommunicator {
         }
 
         if (typeof (volumeThreshold) === "number" ||
-            volumeThreshold === null) {
+            volumeThreshold === null) { // May be NaN
```
howard-stearns, line 1025: move comment to new line 1024?

src/classes/HiFiCommunicator.ts
```diff
@@ -1068,17 +1078,17 @@ export class HiFiCommunicator {
         }
 
         if (typeof (dataJustTransmitted.volumeThreshold) === "number" ||
-            dataJustTransmitted.volumeThreshold === null) {
+            dataJustTransmitted.volumeThreshold === null) { // May be NaN
```
howard-stearns, line 1081: Move comment to new line 1080?

src/classes/HiFiAudioAPIData.ts
howard-stearns: I'm not sure. 
- Part of my confusion is whether it is related to the sentence above. On that issue, maybe it would be clearer to state just the going-forward behavior detached above, and then in the Compatibility Warning note, to specify the additional current behavior along with how that will change. (Maybe also including what the future interpretation will be for null, or zero in the other cases below.)
- Part of my confusion may be about the behavior of 0 vs null here and below.
- And part of my confusion is about what happens if not specified at all. The use of "only fall back to the space volume threshold if the threshold is 'NaN'" seems to say that it has to be explicitly set, but then what would meaningfully happen if not set?

sabrina-shanman: I have updated the compatibility warnings and preceding text and added some details.

howard-stearns (approved):

sabrina-shanman:
Thanks for the review! I am holding off on merging this for the moment so I can do some more testing, to make sure the code still works with the new changes.

sabrina-shanman:
The code appears to still be working, so I will merge this.

PR #184, opened 2021-10-04
HIFI-923 - Allow multiple mixer sessions.

roxanneskelly:
HiFi-923 - Allow multiple mixer sessions.
HIFI-924 - Transition handling.

https://highfidelity.atlassian.net/wiki/spaces/AudioNet/pages/2051309595/Mixer+Transitions
This PR does the following:
* Adds a second "secondaryMixerSession" to the HiFiCommunicator for connection to transition target mixer.
* Adds transition negotiation logic to perform transition as in the above spec.
* sends current user data via audionet.init when reconnecting.

### DIFF
--- a/src/classes/HiFiCommunicator.ts
+++ b/src/classes/HiFiCommunicator.ts
@@ -945,7 +945,7 @@ export class HiFiCommunicator {
         this._inputAudioMediaStreamIsStereo = isStereo;
         let retval = await this._mixerSession.setRAVIInputAudio(newInputAudioMediaStream, isStereo);    
         if (retval) {
-            retval = retval &&  await this._secondaryMixerSession.setRAVIInputAudio(newInputAudioMediaStream, isStereo);
+            retval = await this._secondaryMixerSession.setRAVIInputAudio(newInputAudioMediaStream, isStereo);
         }
         if (!retval){
             HiFiLogger.warn(`Error trying to setRAVIInputAudio on this._mixerSession`);
--- a/src/classes/HiFiCommunicator.ts
+++ b/src/classes/HiFiCommunicator.ts
@@ -793,7 +793,7 @@ export class HiFiCommunicator {
         // change the current state to the new state and call the user's handler.
         if (newState !== this._currentHiFiConnectionState) {
             this._currentHiFiConnectionState = newState;
-            if (!secondary && this.onConnectionStateChanged) {
+            if (this.onConnectionStateChanged) {
                 this.onConnectionStateChanged(this._currentHiFiConnectionState, message);
             }
         }
@@ -1416,7 +1416,7 @@ export class HiFiCommunicator {
         }
 
         let timeoutPerConnectionAttempt = this._connectionRetryAndTimeoutConfig.timeoutPerConnectionAttemptMS;
-        // Kick off the connection attempt. This returns a boolean, but any actual success or failure
+        // Kick off the connection attempt. Any actual success or failure
         // gets handled by the _manageConnection callback handler. (Note that calls to our _connectToHiFiMixer()
         // method get handled entirely by callback-initiated retry code, so we should never get here unless
         // a callback asked us to do it.)
--- a/src/classes/HiFiCommunicator.ts
+++ b/src/classes/HiFiCommunicator.ts
@@ -611,7 +611,7 @@ export class HiFiCommunicator {
                  * They should always kick off a new connection attempt, UNLESS the current state is already "Connected".
                  */
                 if (this._currentHiFiConnectionState !== HiFiConnectionStates.Connected) {
-                    HiFiLogger.log(`_manageConnection: kicking off connection attempt.`);
+                    HiFiLogger.log(`_manageConnection: kicking off connection attempt: ${newState}.`);
                     this._updateStateAndCallUserStateChangeHandler(newState, message, secondary);
                     this._connectToHiFiMixer();
                 } else {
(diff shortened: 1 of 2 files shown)


src/classes/HiFiCommunicator.ts
```diff
@@ -309,6 +333,9 @@ export class HiFiCommunicator {
         onMuteChanged?: OnMuteChangedCallback,
         connectionRetryAndTimeoutConfig?: ConnectionRetryAndTimeoutConfig
     } = {}) {
+        
+        this._visitId = RaviUtils.createUUID();
```
roxanneskelly, line 337: visitid has to be the same across multiple connections, so we cannot use the ravi session id

src/classes/HiFiMixerSession.ts
```diff
@@ -1317,4 +1398,104 @@ export class HiFiMixerSession {
...
+        // send deleted peer notifications for any peers that didn't make it over to the new mixer
+        // and send new peer data notifications for all peers that are on the new mixer.
+        let allDeletedUserData: Array<ReceivedHiFiAudioAPIData> = [];
+        let allNewUserData: Array<ReceivedHiFiAudioAPIData> = [];
+            
+
+        let oldPeers: any = {};
+        for (const mixerPeerKey of Object.keys(this._mixerPeerKeyToStateCacheDict)) {
+            oldPeers[this._mixerPeerKeyToStateCacheDict[mixerPeerKey].hashedVisitID] = this._mixerPeerKeyToStateCacheDict[mixerPeerKey];
+        }
```
AndrewMeadows, line 1476: There is a double map lookup inside the loop above.  Although it takes more lines to write it I think the version below is faster.  That is, it avoids one whole map lookup.
```
for (const key of Object.keys(this._mixerPeerKeyToStateCacheDict)) {
    let value = this._mixerPeerKeyToStateCacheDict[key];
    oldPeers[value.hashedVisitId] = value;
}
```

src/classes/HiFiMixerSession.ts
```diff
@@ -1317,4 +1398,104 @@ export class HiFiMixerSession {
...
+        for (const mixerPeerKey of Object.keys(secondarySession._mixerPeerKeyToStateCacheDict)) {
+            delete oldPeers[secondarySession._mixerPeerKeyToStateCacheDict[mixerPeerKey].hashedVisitID];
+            allNewUserData.push(secondarySession._mixerPeerKeyToStateCacheDict[mixerPeerKey]);
+        }
```
AndrewMeadows, line 1480: Similarly there is a double map lookup in the `for` loop above which could be optimized out.

src/classes/HiFiMixerSession.ts
```diff
@@ -1317,4 +1398,104 @@ export class HiFiMixerSession {
...
+        
+        for (const hashedVisitID of Object.keys(oldPeers)) {
+            let deletedUserData = new ReceivedHiFiAudioAPIData({
+                hashedVisitID: hashedVisitID
+            });
+            if (oldPeers[hashedVisitID].providedUserID) {
+                deletedUserData.providedUserID = oldPeers[hashedVisitID].providedUserID;
+            }
```
AndrewMeadows, line 1488: Similarly this would be faster:
```
let username = oldPeers[hashedVisitID].providedUserID;
if (username) {
    deletedUserData.providedUserID = username;
}
```

AndrewMeadows (changes requested):
It looks functional except for one mysterious no-op `else{}` case.

src/classes/HiFiCommunicator.ts
```diff
@@ -887,10 +941,13 @@ export class HiFiCommunicator {
...
      */
     async setInputAudioMediaStream(newInputAudioMediaStream: MediaStream, isStereo: boolean = false): Promise<boolean> {
-        const retval = await this._mixerSession.setRAVIInputAudio(newInputAudioMediaStream, isStereo);
+        this._inputAudioMediaStream = newInputAudioMediaStream;
+        this._inputAudioMediaStreamIsStereo = isStereo;
+        let retval = await this._mixerSession.setRAVIInputAudio(newInputAudioMediaStream, isStereo);    
         if (retval) {
-            this._inputAudioMediaStream = newInputAudioMediaStream;
-        } else {
+            retval = retval &&  await this._secondaryMixerSession.setRAVIInputAudio(newInputAudioMediaStream, isStereo);
```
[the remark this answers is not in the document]
roxanneskelly: fixing

src/classes/HiFiMixerSession.ts
```diff
@@ -313,6 +336,17 @@ export class HiFiMixerSession {
...
+        this.onTransition = onTransition;
+        this.onReadyForTransition = onReadyForTransition;
+        this._inputAudioMediaStream = undefined;
+        this._inputAudioMediaStreamIsStereo = false;
+        this._currentDataToTransmitToServer = new HiFiAudioAPIData();
+        if (visitId) {
+            this._visitId = visitId;
+        } else {
+            RaviUtils.createUUID();
+        }
```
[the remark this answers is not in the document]
roxanneskelly: oh, that was left-over code from earlier dev.  will remove.  good eye.

src/classes/HiFiMixerSession.ts
```diff
@@ -1058,11 +1118,28 @@ export class HiFiMixerSession {
...
+        let streamController = this._raviSession.getStreamController();
+
+        if (streamController && streamController.getAudioStream()) {
+            raviAudioTracks = streamController.getAudioStream().getAudioTracks();
+        }
+
+        if (raviAudioTracks) {
+            HiFiLogger.log(`Removing this._outputAudioMediaStream tracks to the current RAVI audio tracks.`);
+            raviAudioTracks.forEach(f => this._outputAudioMediaStream.removeTrack(f));
         }
```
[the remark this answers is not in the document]
roxanneskelly: fixing

src/classes/HiFiMixerSession.ts
roxanneskelly: fixing

src/classes/HiFiMixerSession.ts
roxanneskelly: fixing

src/classes/HiFiMixerSession.ts
roxanneskelly: fixing

AndrewMeadows (approved):

PR #193, opened 2021-11-18
Revert "Merge pull request #184 from roxanneskelly/hifi923"

roxanneskelly:
This reverts commit e13584daa8019013224a54035b1b42a7d4612663, reversing
changes made to e2f1a279a6878c693c46d6b6713d6092efe73390.

### DIFF
--- a/src/classes/HiFiMixerSession.ts
+++ b/src/classes/HiFiMixerSession.ts
@@ -18,6 +18,18 @@ import { RaviSignalingConnection, RaviSignalingStates } from "../libravi/RaviSig
 import { Diagnostics } from "../diagnostics/diagnostics";
 import pako from 'pako'
 
+const isBrowser = typeof window !== 'undefined';
+// Since we're initializing a MediaStream, we need to
+// do it using the correct cross-platform class (node or browser)
+let xMediaStream = isBrowser && window.MediaStream;
+if (!isBrowser) {
+    try {
+        xMediaStream = require('wrtc').MediaStream;
+    } catch (e) {
+        ; // Remains falsey. Don't report, don't log
+    }
+}
+
 const INIT_TIMEOUT_MS = 5000;
 const PERSONAL_VOLUME_ADJUST_TIMEOUT_MS = 5000;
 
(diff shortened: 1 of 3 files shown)


src/classes/HiFiMixerSession.ts
```diff
@@ -755,7 +716,6 @@ export class HiFiMixerSession {
             this._raviSession.addStateChangeHandler(this.onRAVISessionStateChanged);
         })
         .catch((error) => {
-            HiFiLogger.warn(error);
```
AndrewMeadows, line 758: This line seems useful.

src/classes/HiFiMixerSession.ts
```diff
@@ -831,11 +790,7 @@ export class HiFiMixerSession {
                 HiFiLogger.warn(`Couldn't set input audio on _raviSession.streamController: No \`streamController\`!`);
                 retval = false;
             } else {
-                if (inputAudioMediaStream) {
-                    streamController.setInputAudio(inputAudioMediaStream.clone(), isStereo);
-                } else {
-                    streamController.setInputAudio(null, isStereo);
-                }
+                streamController.setInputAudio(inputAudioMediaStream, isStereo);
```
AndrewMeadows, line 793: Are we sure we want to lose this `clone()` behavior on the `inputAudioMediaStream`?

src/classes/HiFiMixerSession.ts
```diff
@@ -866,7 +820,6 @@ export class HiFiMixerSession {
                     // If we haven't already connected, it'll just pick up the right stereo value when we
                     // call it the first time.
                     this._inputAudioMediaStreamIsStereo = isStereo;
-                    this._inputAudioMediaStream = inputAudioMediaStream;
```
AndrewMeadows, line 869: Are we sure we want to lose this line?

AndrewMeadows (approved):
I assume this is a pure revert operation.  I spotted a few changes that maybe we would want to keep, but I'm not 100% sure of any of them: I didn't audit how these variables are used.

